PR #7, opened 2015-10-25
[REF] Support of 'noupdate' in check 'duplicate-xml-record-id'

moylop260:
Many times we need duplicate a xml_id one with noupdate="1" and another one
with noupdate="0"
Fix OCA/maintainer-quality-tools#276

### DIFF
--- a/pylint_odoo/misc.py
+++ b/pylint_odoo/misc.py
@@ -205,5 +205,10 @@ class WrapperModuleChecker(BaseChecker):
                 else [self.module, record.get('id')]
             if module and xml_module != module:
                 continue
-            xml_ids.append(xml_module + '.' + xml_id)
+            # Support case where use two xml_id:
+            #  1) With noupdate="1"
+            #  2) With noupdate="0"
+            noupdate = "noupdate=" + record.getparent().get('noupdate', '0')
+            xml_ids.append(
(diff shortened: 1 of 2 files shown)


guewen:
:+1: Thanks!

pylint_odoo/misc.py
```diff
@@ -205,5 +205,10 @@ def get_xml_record_ids(self, xml_file, module=None):
                 else [self.module, record.get('id')]
             if module and xml_module != module:
                 continue
-            xml_ids.append(xml_module + '.' + xml_id)
+            # Support case where use two xml_id:
```
pedrobaeza: s/use/using

pedrobaeza:
Little nitpicking, but :+1:

pylint_odoo/misc.py
moylop260: @pedrobaeza 
Thank you
Fixed!

moylop260:
Fixed typo

dreispt:
:+1: for this.

This check should also be done in CSVs, at least for Security ACL files following the standard columns structure: I recently saw a bug fix where the lines where duplicated but the author forgot to change the ID of the second line, so it overwrites the former permission and the module doesn't work.

moylop260:
@dreispt 
Good point
Could you create a new issue to have a ToDo list?

PR #10, opened 2015-10-26
[ADD] Add check 'deprecated-openerp-xml-node'

moylop260:
It don't is deprecated now but this check can be used in the future.
Then we can have this check in pylint-odoo (for use it in other projects) but we can wait to enabled in oca/MQT

### DIFF
--- a/pylint_odoo/test/main.py
+++ b/pylint_odoo/test/main.py
@@ -8,7 +8,7 @@ from pylint.lint import Run
 from pylint_odoo import misc
 
 
-EXPECTED_ERRORS = 47
+EXPECTED_ERRORS = 52
 
 
 class MainTest(unittest.TestCase):
--- a/pylint_odoo/checkers/modules_odoo.py
+++ b/pylint_odoo/checkers/modules_odoo.py
@@ -43,13 +43,17 @@ ODOO_MSGS = {
         'javascript-lint',
         settings.DESC_DFLT
     ),
+    'W%d04' % settings.BASE_OMODULE_ID: (
+        '%s:%d Deprecated <openerp> xml node',
+        'deprecated-openerp-xml-node',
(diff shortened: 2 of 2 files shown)


pedrobaeza:
But is this deprecated now? Or in the future?

moylop260:
@pedrobaeza 
It don't is deprecated now but this check can use be used in the future.
Then we can have this check in pylint-odoo (for use it in my projects) but we can wait to enabled in oca/MQT

pedrobaeza:
OK then :+1:

dreispt:
:+1:

dreispt:
Idea: add a check for simple `<data>` element (with no `update` attribute). They are redundant in >=8.0 and we could make a warning for that.

pylint_odoo/test/main.py
```diff
@@ -8,7 +8,7 @@
 from pylint_odoo import misc
 
 
-EXPECTED_ERRORS = 46
+EXPECTED_ERRORS = 50
```
dreispt: Suggestion - to avoid the annoying PR conflicts you could write something like this: 
```python
EXPECTED_ERRORS = 46
EXPECTED_ERRORS += 4
```

dreispt:
Rebase :recycle:

pylint_odoo/test/main.py
moylop260: @dreispt 
Good point
Thank you!

moylop260:
> Idea: add a check for simple <data> element (with no update attribute). They are redundant in >=8.0 and we could make a warning for that.

@dreispt 
Good idea
New issue?

moylop260:
rebased

moylop260:
ping

pedrobaeza:
We can merge this one

PR #11, opened 2015-10-26
[IMP] Add check 'create-user-without-reset-password'

moylop260:
Many times we have next traceback in our customer builds projects:
```bash
2015-10-08 20:32:43,139 232 ERROR openerp_test openerp.addons.base.ir.ir_mail_server: Mail delivery failed via SMTP server 'localhost'.
error: 111
Connection refused
2015-10-08 20:32:43,139 232 ERROR openerp_test openerp.addons.mail.mail_mail: failed sending mail.mail 21
File "addons/mail/mail_mail.py", line 304, in send
context=context)
File "openerp/api.py", line 241, in wrapper
return old_api(self, *args, **kwargs)
File "openerp/addons/base/ir/ir_mail_server.py", line 503, in send_email
raise MailDeliveryException(_("Mail Delivery Failed"), msg)
MailDeliveryException: ('Mail Delivery Failed', u"Mail delivery failed via SMTP server 'localhost'.\nerror: 111\nConnection refused")
```

The module [auth_signup](https://github.com/odoo/odoo/blob/c4844675b83e299b5d9f81418142c61688f4546a/addons/auth_signup/res_users.py#L303-L306) avoid this traceback with a context when you create users.

```xml
...
 <record model="res.users" id="external_id" context="{'no_reset_password': True}">
    <field name="name">Test user 1</field>
    <field name="login">test_user1</field>
    ...
...
```

IMHO If we create a user should avoid future traceback with other projects that use `auth_signup` module.

### DIFF
--- a/pylint_odoo/test/main.py
+++ b/pylint_odoo/test/main.py
@@ -8,7 +8,7 @@ from pylint.lint import Run
 from pylint_odoo import misc
 
 
-EXPECTED_ERRORS = 46
+EXPECTED_ERRORS = 47
 
 
 class MainTest(unittest.TestCase):
--- a/.coveragerc
+++ b/.coveragerc
@@ -15,3 +15,5 @@ exclude_lines =
     if __name__ == '__main__':
     # Don't complain about uninstalled packages
     except OSError as oserr:
+    except NameError:
+    except SyntaxError:
--- a/pylint_odoo/checkers/modules_odoo.py
+++ b/pylint_odoo/checkers/modules_odoo.py
@@ -43,6 +43,13 @@ ODOO_MSGS = {
         'javascript-lint',
         settings.DESC_DFLT
     ),
+    'W%d05' % settings.BASE_OMODULE_ID: (
+        '%s:%d record res.users without '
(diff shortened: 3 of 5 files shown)


pylint_odoo/test/main.py
```diff
@@ -8,7 +8,7 @@
 from pylint_odoo import misc
 
 
-EXPECTED_ERRORS = 46
+EXPECTED_ERRORS = 47
```
pedrobaeza: In other PR you change also this number

pedrobaeza:
I don't understand the rationale of this change

pylint_odoo/test/main.py
moylop260: Yes, all PR it will diverge it later merge one I will rebase it

moylop260:
> I don't understand the rationale of this change

@pedrobaeza 
You are right I missed add a description to this PR.
Could you check main description, please

pedrobaeza:
Thanks for the explanation, and indeed is a good practise and we can check it with this.

Good work! :+1:

dreispt:
:+1:

PR #17, opened 2015-12-18
[IMP] pylint_odoo: Add new check wizard-bad-defined

moylop260:
- The wizard class should be inside wizards/ path.

### DIFF
--- a/pylint_odoo/checkers/no_modules.py
+++ b/pylint_odoo/checkers/no_modules.py
@@ -136,6 +136,11 @@ ODOO_MSGS = {
         'license-allowed',
         settings.DESC_DFLT
     ),
+    'C%d07' % settings.BASE_NOMODULE_ID: (
+        'The wizard class should be inside wizards/ path.',
+        'wizard-bad-defined',
+        settings.DESC_DFLT
+    ),
 }
 
 DFTL_MANIFEST_REQUIRED_KEYS = ['license']
@@ -152,6 +157,10 @@ DFTL_ATTRIBUTE_DEPRECATED = [
 DFTL_METHOD_REQUIRED_SUPER = [
     'create', 'write', 'read', 'unlink', 'copy',
     'setUp', 'tearDown', 'default_get',
+]
+DFTL_WIZARD_CLASSES = [
+    'TransientModel',
+    'osv_memory',
 ]
 
 
@@ -201,6 +210,13 @@ class NoModuleChecker(BaseChecker):
             'metavar': '<comma separated values>',
             'default': DFTL_METHOD_REQUIRED_SUPER,
             'help': 'List of methods where call to `super` is required.' +
+                    'separated by a comma.'
+        }),
+        ('wizard_classes', {
+            'type': 'csv',
+            'metavar': '<comma separated values>',
+            'default': DFTL_WIZARD_CLASSES,
+            'help': 'Name of classes base used to define a wizard in odoo.' +
                     'separated by a comma.'
         }),
     )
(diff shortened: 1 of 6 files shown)


moylop260:
Ping

dreispt:
LGTM :+1:

guewen:
:-1: this check goes too far I think. TransientModels are still models and even if most of the time they are used as wizards, I think we should not forbid to create them elsewhere than in the wizards directory. It should be a convention, not a rule.

Personally, I don't understand the rationale of putting them in a `wizards` directory, it is a legacy of the old versions where wizards were not models. It makes often more sense to have them grouped with the models they are acting on (e.g. in the file where I have `queue.job`, I also have the TransientModel that re-enqueue jobs).

moylop260:
@guewen 
You should avoid discuss guidelines in each PR.
You should use the correct place to change guidelines:
https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#directories

If we search your proposal to change guidelines we have 0 [pr@guewen](https://github.com/OCA/maintainer-tools/pulls/guewen)

But in each new PR you can change it.

Could you create a new PR with all your changes to MT, please?
This could allow us work more efficient.

guewen:
My message had no intent to change any guidelines. My point is that some
conventions such as pep8 should be enforced and must raise build errors
when violated, wherever others should be left at the appreciation of the
developers and the reviewers. I place this "wizard" convention in the
latter without any doubts.

As a reviewer, I would not refuse a PR because it has TransientModels
outside of the wizard folder. The barrier to start developing in the OCA is
already too high for newcomers, please keep what we could call the "Must
have conventions" as build errors and the "better but not mandatory" simply
as statements in the guidelines.

IMO it would be an error to create build errors for any single rule of the
guidelines for the reason above, and that's why I gave a :-1:

moylop260:
I'm talking about:

> Personally, I don't understand the rationale of putting them in a `wizards` directory, it is a legacy of the old versions where wizards were not models. It often makes more sense to have them grouped with the models they are acting on (e.g. in the file where I have `queue.job`, I also have the TransientModel that re-enqueue jobs).

You are giving a reason to avoid use the guideline defined to `wizards` directories.
But this project don't is a correct place to add this comment.
You should add a new PR to MT to change this guidelines.


FYI the focus of this project is auto-review the guidelines of MT.
All features of this project has the same focus, example:
 - CamelCase
 - Missing Readme
 - Deprecated keys
 - License
 - coding
 - Duplicated xml id
 - vim comment
And other ones documented MT guidelines..

Many reviewers are creating manually comments of follow MT guidelines, but we have this project to this proposal, if you don't want a guidelines you can remove it from MT.
If a guidelines exists in MT then we are working to added it to this plugin too.

If you don't want auto-block a PR, we can enabled in MQT in `beta messages`.

With your criteria, all PR of this project will be rejected.

guewen:
> Personally, I don't understand the rationale of putting them in a wizards directory, it is a legacy of the old versions where wizards were not models. It often makes more sense to have them grouped with the models they are acting on (e.g. in the file where I have queue.job, I also have the TransientModel that re-enqueue jobs).
> 
> You are giving a reason to avoid use the guideline defined to wizards directories.
> But this project don't is a correct place to add this comment.
> You should add a new PR to MT to change this guidelines.

It was just a personal digression. I didn't asked to change the convention.


> Many reviewers are creating manually comments of follow MT guidelines, but we have this project to this proposal, if you don't want a guidelines you can remove it from MT.
> If a guidelines exists in MT then we are working to added it to this plugin too.

This is where I disagree. As I told before, I think there are rules that must inevitably be enforced and others that must stay in the MT guidelines as convention but should not be strictly enforced.

> 
> If you don't want auto-block a PR, we can enabled in MQT in beta messages.

Nobody will read the beta messages if we have dozen of lines.

> 
> With your criteria, all PR of this project will be rejected.

Why? 
The examples you cite above (CamelCase, Missing Readme, deprecated keys, ...) seems reasonable to check to me, it does not mean that every single line of the MT conventions should be strictly checked.

Some conventions are okay to be broken, when you understand the reason, and it can only be appreciated by humans, not by an automated check. If you want examples, I can say `res.config.settings` which are not wizards per se; if you know the `account_advanced_reconcile` addon, there are also [TransientModels](https://github.com/OCA/bank-statement-reconcile/blob/8.0/account_advanced_reconcile/advanced_reconciliation.py#L25) which are used for their Transient property but are not *wizards*.

guewen:
TLDR: `TransientModel` are not always wizards.

moylop260:
@guewen 
Thanks for clarify
Your last example is very good!

NOTE: His personal digression made some noise

PR #18, opened 2016-04-05
[IMP] pylint_odoo: Add check for manifest-version-format

lescobarvx:
[IMP] pylint_odoo: Add check for manifest-version-format


moylop260:
Could you run 
[messages2rst](https://github.com/OCA/pylint-odoo/blob/e12d096418f11172de8152e1413c9cdc7c7cbaf2/pylint_odoo/__init__.py#L35) method and modified the README.rst file with your new check, please?

pylint_odoo/test/main.py
```diff
@@ -8,7 +8,7 @@
 from pylint_odoo import misc
 
 
-EXPECTED_ERRORS = 52
+EXPECTED_ERRORS = 54
```
pedrobaeza: As commented other times, it would be better to add a line like this:
```
EXPECTED_ERRORS += 2  # Errors due to C8106: Wrong version format
```
This way, multiple PRs don't get affected each other and the number seems less random.

pedrobaeza:
Thanks!

:+1:

lescobarvx:
@pedrobaeza 

You're welcome!!

moylop260:
:+1:

pylint_odoo/checkers/no_modules.py
```diff
@@ -261,6 +274,15 @@ def visit_dict(self, node):
                 self.add_message('license-allowed',
                                  node=node, args=(license,))
 
+            # Check version format
+            version_format = manifest_dict.get('version', None)
```
moylop260: Change the code:
```dif
-            version_format = manifest_dict.get('version', None)
+           version_format = manifest_dict.get('version', '')
```
To avoid follow error:
screen shot 2016-04-06 at 9 08 52 pm

moylop260:
@guewen @max3903 @dreispt @lepistone your feedback is welcome

pylint_odoo/test/main.py
```diff
@@ -9,6 +9,7 @@
 
 
 EXPECTED_ERRORS = 52
+EXPECTED_ERRORS += 2  # Errors due to C8106: Wrong version format
```
max3903: @lescobarvx why not `EXPECTED_ERRORS = 54` in one line and add the comment in the commit message?

max3903:
CLA is OK.

pylint_odoo/test/main.py
lescobarvx: @max3903 
@pedrobaeza indicated me to do so because that way multiple PRs don't get affected each other and the number seems less random.

max3903: @lescobarvx ok, thanks.

moylop260: First version was created with 54

But @pedrobaeza recommend us use with +2

screen shot 2016-04-06 at 11 40 44 pm

moylop260:
@max3903 
Thanks for feedback
Could you add your feedback [here](https://github.com/OCA/maintainer-quality-tools/pull/325) too=?

pylint_odoo/test/main.py
pedrobaeza: Yeah, is more easy to follow the total number of expected errors and to avoid conflicts if multiple PRs are done simultaneously.

max3903: @pedrobaeza What do you propose to avoid having this after couple contributions?
```
EXPECTED_ERRORS = 52
EXPECTED_ERRORS += 2  # Errors due to C8106: Wrong version format
EXPECTED_ERRORS += 3  # Errors due to XXXX
EXPECTED_ERRORS += 4  # Errors due to YYYY
EXPECTED_ERRORS += 5  # Errors due to ZZZZ
```

max3903: :o) Ok, but at some point (when there is no active PR), we should reset that to a single line.

Let's merge it and move forward.

pedrobaeza: Well, I prefer to let unfolded. As said, this allows to identify how many errors of each type you have introduced. If you have a global counter, you need to check each of them for a global number, and it more prone to put the given number by the test than the real expected one.

pedrobaeza:
@moylop260, should we make a new version?

moylop260:
> @moylop260, should we make a new version?

Yes

pedrobaeza:
As Luis is proposing other changes, we can wait for them.

PR #20, opened 2016-04-15
[IMP] Add check Duplicate field in xml record

lescobarvx:
Check same field assigned 2 times in xml record.

### DIFF
--- a/pylint_odoo/test_repo/broken_module/model_view_odoo2.xml
+++ b/pylint_odoo/test_repo/broken_module/model_view_odoo2.xml
@@ -1,0 +1,76 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+    <data>
+        <!-- duplicate record field "name" -->
+        <record id="view_model_form" model="ir.ui.view">
+            <field name="name">view.model.form</field>
+            <field name="model">test.model</field>
+            <field name="arch" type="xml">
+                <tree string="Test model">
+                    <field name="name"></field>
+                </tree>
+            </field>
+            <field name="name">view.model.form4</field>
+        </record>
+
+        <!-- inherit view with duplicated is not checked -->
+        <record id="view_model_form2" model="ir.ui.view">
+            <field name="name">view.model.form</field>
+            <field name="model">test.model</field>
+            <field name="inherit_id" ref="id_category_list"/>
+            <field name="arch" type="xml">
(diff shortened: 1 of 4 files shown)


lescobarvx:
@moylop260 @pedrobaeza @max3903 
Please your feedback

pedrobaeza:
You should check or at least put as known issues that duplicated fields across inherited views are not taken into account.

lescobarvx:
@pedrobaeza 
the change is ready where if it finds an inherited view does not take into account

pedrobaeza:
So it's working with inherited views or not?

lescobarvx:
@pedrobaeza 
It doesn't work with inherited views because I haven't a viable solution to check them at this time

pedrobaeza:
OK, then you should mention it in the description of the warning.

pedrobaeza:
:+1:

lescobarvx:
@pedrobaeza 
Ready, I put an important note in the docstring of _check_duplicate_xml_fields method indicating that this check does not work with inherited views, that's OK?

pedrobaeza:
Yeah, I see. You already have my :+1:.

Thanks for the work.

lescobarvx:
@pedrobaeza 
you're welcome

moylop260:
Please, add a test case with inherit

pylint_odoo/test_repo/broken_module/model_view_odoo2.xml
```diff
@@ -0,0 +1,61 @@
...
+                </tree>
+            </field>
+            <field name="name">view.model.form4</field>
+        </record>
+
+        <!-- inherit view is not checked -->
+        <record id="view_model_form2" model="ir.ui.view">
+            <field name="name">view.model.form</field>
+            <field name="model">test.model</field>
+            <field name="inherit_id" ref="id_category_list"/>
```
lescobarvx: @moylop260 
Ready, added test case of inherited views.

moylop260: Duplicated and non duplicated with inherit here too pleaseplease

lescobarvx:
@moylop260 @pedrobaeza 
Ready, 2 test case for inherited views: with duplicate fields and without duplicate fields

moylop260:
Thanks
:+1:

PR #21, opened 2016-04-16
[IMP] test: Add tox for multi-environment and add compatiblity with latest version

lescobarvx:


lescobarvx:
@moylop260 @pedrobaeza @max3903 
Your feedback please

.travis.yml
```diff
@@ -30,13 +36,15 @@ install:
     - pip install --upgrade pyopenssl ndg-httpsclient pyasn1
     - pip install coveralls flake8
 
+    # Install tox
+    - pip install tox
```
pedrobaeza: Why do you need this?

requirements.txt
```diff
@@ -1,6 +1,6 @@
-pylint==1.4.4
+-e git+git://github.com/PyCQA/pylint.git@master#egg=pylint
```
pedrobaeza: You shouldn't use last master version, but stable one.

.travis.yml
moylop260: If you run locally:
```python
cd pylint-odoo
tox
```
You will be tested 2 virtual environments of pylint-odoo
 - pylint1.4
 - pylint2.0

tox is the most library used in python to multi-environment cases.

requirements.txt
moylop260: Yes, this is a known issue because currently we don't have a release in pylint, but now I have created the question to create a release: https://github.com/PyCQA/pylint/issues/874#issuecomment-210949485

moylop260:
Rebased.

This PR add compatibility with new versions of pylint and astroid.
We are using in requirements.txt the current stable version, but developing for old and new version with small changes.

dreispt:
LGTM :+1:

moylop260:
@nhomar @pedrobaeza your feedback is welcome!

moylop260:
👍

PR #24, opened 2016-04-19
[IMP] Add Check tabs identation instead of four spaces

lescobarvx:
[IMP] Add Check tabs identation instead of four spaces

### DIFF
--- a/pylint_odoo/checkers/modules_odoo.py
+++ b/pylint_odoo/checkers/modules_odoo.py
@@ -67,11 +67,17 @@ ODOO_MSGS = {
         'redundant-modulename-xml',
         settings.DESC_DFLT
     ),
+    'W%d10' % settings.BASE_OMODULE_ID: (
+        '%s:%s Use wrong tabs indentation instead of four spaces',
+        'wrong-tabs-instead-of-spaces',
+        settings.DESC_DFLT
+    ),
 }
 
 
 DFTL_README_TMPL_URL = 'https://github.com/OCA/maintainer-tools' + \
     '/blob/master/template/module/README.rst'
+DFTL_EXTFILES_TO_LINT = ['xml', 'csv', 'po', 'js', 'mako']
 
 
 class ModuleChecker(misc.WrapperModuleChecker):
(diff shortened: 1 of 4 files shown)


pylint_odoo/checkers/modules_odoo.py
```diff
@@ -149,6 +161,24 @@ def _check_duplicate_id_csv(self):
...
+                 add list of errors in self.msg_args
+        """
+        self.msg_args = []
+        for type_file in self.config.extfiles_to_lint:
+            for ext_file in self.filter_files_ext(type_file, relpath=False):
+                countline = 0
+                with open(ext_file, 'rb') as fp:
+                    for line in fp:
+                        countline += 1
+                        line and line[0] == '\t' and self.msg_args.append(
```
pedrobaeza: You should check for any tab inside the line. Not only at the beginning. You can have four spaces, then a tab, then other four spaces...

lescobarvx:
@pedrobaeza 
The change is ready

pedrobaeza:
:+1:

pylint_odoo/checkers/modules_odoo.py
```diff
@@ -149,6 +161,24 @@ def _check_duplicate_id_csv(self):
...
+                 add list of errors in self.msg_args
+        """
+        self.msg_args = []
+        for type_file in self.config.extfiles_to_lint:
+            for ext_file in self.filter_files_ext(type_file, relpath=False):
+                countline = 0
+                with open(ext_file, 'rb') as fp:
+                    for line in fp:
+                        countline += 1
+                        line and '\t' in line and self.msg_args.append(
```
moylop260: We can have a good `tab` in a line.
Add follow test:
```xml
    <field name="field">My value with tabs \t</field>
```

Maybe, if you use a `lstrip('\t')` and It compare with the original string you can have a better approach.

moylop260: Add a validation to use four spaces instead of 2 spaces of 3 spaces...

pylint_odoo/checkers/modules_odoo.py
```diff
@@ -149,6 +161,24 @@ def _check_duplicate_id_csv(self):
...
         return True
 
+    def _check_tabs_instead_of_spaces(self):
+        """Check tabs character instead of four spaces.
+        :return: False if exists errors and
+                 add list of errors in self.msg_args
+        """
+        self.msg_args = []
+        for type_file in self.config.extfiles_to_lint:
+            for ext_file in self.filter_files_ext(type_file, relpath=False):
```
moylop260: If is a js from lib folder we need jump it (Don't check it)
Because generally are external libraries that we don't wish change it

dreispt:
This sounds like a generic Python lint.
Would it be worth proposing it to pylint?

moylop260:
Python have this check currently
This one if for other type of technicall files as xml, js...

moylop260:
Rebased

nhomar:
:+1:

moylop260:
👍

PR #25, opened 2016-04-19
[IMP] Add pylint check Redundant module name reference in xml_id

lescobarvx:
[IMP] Add pylint check Redundant module name reference in xml_id

### DIFF
--- a/pylint_odoo/checkers/modules_odoo.py
+++ b/pylint_odoo/checkers/modules_odoo.py
@@ -63,7 +63,7 @@ ODOO_MSGS = {
         settings.DESC_DFLT
     ),
     'W%d09' % settings.BASE_OMODULE_ID: (
-        'Redundant name module reference in xml_ids "%s" of "%s" file',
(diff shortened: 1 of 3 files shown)


moylop260:
👍

dreispt:
At least in 8.0, the `<report>` elements were required to use XML Ids with `module.name`.
Example: https://github.com/odoo/odoo/blob/9.0/addons/account_extra_reports/data/account_report.xml

moylop260:
The `id="action_report_journal"`  is right a wrong redundant could be `id="account_extra_reports.action_report_journal"`
However, this check just is used in `<record id=...` case

pedrobaeza:
@moylop260, please rebase. 

@dreispt, indeed, this only checks `<record>` tag, not other special tags like `<report>`, `<menuitem>`...

moylop260:
I'm waiting the follow merge https://github.com/OCA/pylint-odoo/pull/26 to fix conflicts here

pedrobaeza:
:+1:

PR #28, opened 2016-05-17
[ADD] invalid-commit: Add new check invalid-commit

moylop260:
- Verify the guideline [never-commit-the-transaction](https://github.com/oca/maintainer-tools/blob/master/CONTRIBUTING.md#never-commit-the-transaction)

### DIFF
--- a/pylint_odoo/checkers/no_modules.py
+++ b/pylint_odoo/checkers/no_modules.py
@@ -107,6 +107,13 @@ ODOO_MSGS = {
         'manifest-author-string',
         settings.DESC_DFLT
     ),
+    'E%d02' % settings.BASE_NOMODULE_ID: (
+        'Use of cr.commit() directly - More info '
+        'https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md'
+        '#never-commit-the-transaction',
+        'invalid-commit',
+        settings.DESC_DFLT
+    ),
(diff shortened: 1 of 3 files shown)


pedrobaeza:
:+1:

lescobarvx:
:+1:

guewen:
Commit could still be valid such as in cron tasks or for performance reasons.
This is also noticed by the guidelines in:

> All cr.commit() calls outside of the server framework from now on must have an explicit comment explaining why they are absolutely necessary, why they are indeed correct, and why they do not break the transactions. Otherwise they can and will be removed!

moylop260:
@guewen 
You are right in some cases we need `cr.commit`
First step: Could you use cr.savepoint? No
Second step: Could you use a duplicated cursor to isolated the current operation (More info [here](https://www.odoo.com/es_ES/forum/ayuda-1/question/how-to-get-a-new-cursor-on-new-api-on-thread-63441))? No
Third step: Then use `cr.commit`, but add an explicit comment explaining... in this comment you can add `# pylint: disable=invalid-commit` to bypass this check.

guewen:
> Third step: Then use cr.commit, but add an explicit comment explaining... in this comment you can add # pylint: disable=invalid-commit to bypass this check.

ok noted, thanks
:+1:

oscarolar:
👍

moylop260:
Somebody, could be merged it?

PR #31, opened 2016-05-17
[REF] javascript-lint: Ignore /lib/ subfolder

moylop260:
- Many external libraries of javascript are used in our odoo modules, but we can't to maintain, then should be without changes.


lescobarvx:
:+1:

oscarolar:
👍

pedrobaeza:
:+1:

PR #33, opened 2016-05-17
[ADD] method-param: Check param method starts with "_param_"

moylop260:
- Add method-compute, method-search, method-inverse that the method starts with `_compute_`, `_search_` and `_inverse_` respectively
 - Based on guideline: https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#field
 - Inspired in @guewen https://github.com/OCA/rma/pull/87#discussion_r62834498
guewen


moylop260:
inspired in @guewen https://github.com/OCA/rma/pull/87#discussion_r62834498

pedrobaeza:
:+1:

cyrilgdn:
:+1:

lescobarvx:
:+1:

moylop260:
Could merge it?

moylop260:
@dufresnedavid your feedback is welcome here too

lmignon:
@moylop260 Is travis will fail if you change the code of a module that does not comply with the new rules?
You must pay attention that if you require to rename an existing method to comply with these new rules, you'll break the API. In this case the new rules can only be applied for Odoo 10 for existing modules.

dufresnedavid:
@moylop260 I agree with @lmignon , travis must not fail for 7.0, 8.0 and 9.0, otherwise, all pull request are likely going to fail. We don't want to rename existing methods only so that pylint is happy.

pedrobaeza:
These will be activated as beta messages, which means that doesn't turn into red Travis builds.

Anyway, don't get confused with this repository. This will only allow to make these checks. Its scope is to extend pylint functionalities to detect this kind of Odoo specific errors. It will be in MQT (maintainer-quality-tools) repo where we must activate them for being used.

dufresnedavid:
@pedrobaeza thanks for the explanation.

:+1:

moylop260:
>  travis must not fail for 7.0, 8.0 and 9.0

I will care that in https://github.com/OCA/maintainer-quality-tools/pull/330

PR #34, opened 2016-05-21
[IMP] pylint_odoo: Enable cProfile print stats and improve performance

moylop260:
- Add a profiling report of top five `pytlint_odoo` methods more time consuming or more times executed.

e.g.
screen shot 2016-05-21 at 2 26 45 pm

 - This report is useful for developers to improve performance.

 - Improve performance, now show minor cProfile values:
screen shot 2016-05-21 at 4 50 56 pm

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,7 +41,7 @@ install:
 
 script:
     - flake8 --exclude=__init__.py .
-    - tox -e $TOXENV
+    - tox -e $TOXENV,profile-stats
 
 after_success:
     - coveralls
--- a/pylint_odoo/misc.py
+++ b/pylint_odoo/misc.py
@@ -69,6 +69,28 @@ class WrapperModuleChecker(BaseChecker):
                 if os.path.isfile(manifest_file):
                     return manifest_file
 
+    def set_ext_files(self):
+        """Create `self.ext_files` dictionary with {extension_file: [files]}
+        """
+        self.ext_files = {}
+        for root, _, filenames in os.walk(self.module_path, followlinks=True):
+            for filename in filenames:
+                fext = os.path.splitext(filename)[1].lower()
+                fname_rel = os.path.relpath(
+                    os.path.join(root, filename), self.module_path)
+                self.ext_files.setdefault(fext, []).append(fname_rel)
+
+    def set_caches(self):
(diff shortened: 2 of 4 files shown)


dreispt:
LGTM +1

moylop260:
ping

dreispt:
You already have my :+1:
This is for self checks, right?
It would be grate to also have this on MQT 😉

moylop260:
@dreispt 
Thanks for you approval
We will wait approvals of other people.

> This is for self checks, right?

Yes, It's

> It would be grate to also have this on MQT 😉

Are you talking about for self-checks of MQT or for the repo modules that using MQT?


FYI I have a similar report for modules that using VX/MQT
screen shot 2016-07-01 at 7 27 57 pm

This is very useful to detect some performance issues early when you have high level of coverage.
With this tool I saw that methods need a refactory.

I have the following PR pending from 5 months ago to add it to OCA/MQT
https://github.com/OCA/server-tools/pull/375

I have other one for pgbadger sql analysis:

screen shot 2016-07-01 at 7 43 12 pm

With this tool I saw that we have cache issues in 8.0 but in 9.0 is fixed:
 - [8.0-get_param-without-cache](https://github.com/odoo/odoo/blob/b9a006c1c02a06cf3cd34828860125e6d6aa3fb4/openerp/addons/base/ir/ir_config_parameter.py#L71)
 - [9.0-get_param-with-cache](https://github.com/odoo/odoo/blob/1d1debf6ae521d766bec10792d50b4835c0f479f/openerp/addons/base/ir/ir_config_parameter.py#L66)

I'm working to have this same tools in OCA/MQT but the process is too slow... Example this PR from 2 months ago.

I hope help to unblock this features in order to enjoy them in OCA repos. (Maybe Belgium's OCA sprint)

dreispt:
My question was:
The purpose of this to check the performance of the self tests, correct?
Can it be used to check the performance of tested modules in MQT builds?

moylop260:
> The purpose of this to check the performance of the self tests, correct?
Yes

> Can it be used to check the performance of tested modules in MQT builds?
yes, with https://github.com/OCA/server-tools/pull/375 plus a small in MQT

osvalr:
this is looking good, :+1:

PR #40, opened 2016-06-24
[IMP] deprecated-module: Add exception for openerp.osv.expression

moylop260:
Fix https://github.com/OCA/maintainer-quality-tools/issues/306 - `from openerp.osv import expression` is a valid import yet.

Output result example:
screen shot 2016-06-24 at 6 13 49 pm

- [x] [REF] tests: Add a dummy package `openerp.osv` to force `deprecated-module` [pylint#362](https://bitbucket.org/logilab/pylint/issues/362/bug-w0402-not-show-fails-if-module-not-is)
- [x] [REF] tests: Add `sys.path` to allow `import openerp.osv`

### DIFF
--- a/pylint_odoo/augmentations/main.py
+++ b/pylint_odoo/augmentations/main.py
@@ -3,6 +3,7 @@ import os
 
 from pylint_plugin_utils import suppress_message
 from pylint.checkers.base import BasicChecker
+from pylint.checkers.imports import ImportsChecker
 from .. import settings
 
 
@@ -14,6 +15,22 @@ def is_manifest_file(node):
     return is_manifest
 
 
+def is_valid_openerp_osv_deprecated(node):
+    """Verify if the node is a import deprecated but valid
+
+    E.g. "from openerp.osv import expression" is valid
+        but "from openerp.osv import {other} is invalid"
+    :returns: True if is a valid import
+    """
+    submodules = [submodule[0] for submodule in node.names]
+    modname = getattr(node, 'modname', [])
+    if len(submodules) == 1 and (
+            submodules[0] == 'openerp.osv.expression' or
+            submodules[0] == 'expression' and 'openerp.osv' in modname):
+        return True
+    return False
+
+
 def apply_augmentations(linter):
     """Apply suppression rules."""
 
@@ -22,3 +39,10 @@ def apply_augmentations(linter):
     discard = hasattr(BasicChecker, 'visit_discard') and \
         BasicChecker.visit_discard or BasicChecker.visit_expr
     suppress_message(linter, discard, 'W0104', is_manifest_file)
+
+    # W0402 - deprecated-module valid openerp.osv.expression
+    discard = ImportsChecker.visit_import
+    suppress_message(linter, discard, 'W0402', is_valid_openerp_osv_deprecated)
+    discard = hasattr(ImportsChecker, 'visit_from') and \
+        ImportsChecker.visit_from or ImportsChecker.visit_importfrom
+    suppress_message(linter, discard, 'W0402', is_valid_openerp_osv_deprecated)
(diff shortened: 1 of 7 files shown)


moylop260:
@StefanRijnhart 
This is the feature requested

StefanRijnhart:
Great work, thanks! :+1:

pedrobaeza:
As there's a valid alias in Odoo (`from openerp.models import expression`), I see no need for this. References: 
8.0 - https://github.com/odoo/odoo/blob/ee3666237af305de7960103b66ab68afa9c894a9/openerp/models.py#L6215
9.0 - https://github.com/odoo/odoo/blob/44d23c371cebf888bc7438a4208f14a3ad789f99/openerp/models.py#L6282

moylop260:
We can use transitive imports to bypass the error.
But we have people that dont like to use them.

Check the following comment of @gurneyalex

https://github.com/OCA/maintainer-quality-tools/issues/306#issuecomment-171592748

Other people want disable the check.

Other people want the check with transitive imports.

IMHO this change allow you use the best option for you without a problem of a false deprecated.

The expression is present from osv in latest master odoo version yet

If we search how odoo team `import` expression... we saw that they aren't use the transitive imports.
They are using osv.expression.

pedrobaeza:
OK, but please include tests for the new method that doesn't lower the coverage.

moylop260:
The test is included...
But in the main description I explained why don't work the coverage is down...

moylop260:
FYI
Im workin on fix https://github.com/PyCQA/pylint/pull/954/files
But we will need wait a pylint release to have it in stable

This section just affect our tests because we don't have other tests that use this method.

Maybe if I force other test just to bypass the case...

Let me see

moylop260:
FYI this PR is ready

moylop260:
ping

moylop260:
ping

moylop260:
Well, 
IMHO We should take the 👍  of @guewen @yvaucher from https://github.com/OCA/maintainer-quality-tools/issues/306 comments too. Then I'll merge it with a big "sorry" (after 2 pings and 17 days ago without answers)

StefanRijnhart:
Is the OCA dying?

yvaucher:
@StefanRijnhart Maybe a part is on vacation? OCA is now quite big. Hard to follow everything.

pedrobaeza:
These are very specific topics that not everyone can follow nor opinate. I was also quite busy these days to confirm that my comments were attended.

dreispt:
@StefanRijnhart most Pylint & MQT PR are proposed by @moylop260 and then reviewed by me and @pedrobaeza 

In this case @pedrobaeza mentioned some concerns about this, and I think he has a point and was waiting for counter-arguments before I expressed an opinion.
That was enough to stall thhe PR.

StefanRijnhart:
Oh, it's not just this particular PR. Most projects I visit are disfunctional.

dreispt:
@StefanRijnhart Disfunctional in what way? Stalled PRs?

gurneyalex:
@StefanRijnhart I've been really under the water during the 1st semester 2016. I'm recovering and slowly processing the github messages backlog that has stacked up high in my mailbox... 

May I suggest you raise the topic on the contributors mailing list, which is a better forum than github for such a broad discussion?

StefanRijnhart:
On 08/03/2016 09:14 AM, Alexandre Fayolle wrote:
>
> @StefanRijnhart <https://github.com/StefanRijnhart> I've been really
> under the water during the 1st semester 2016. I'm recovering and
> slowly processing the github messages backlog that has stacked up high
> in my mailbox...
>

Ah, I was wondering about your absense. Sorry to hear you were not well.
Glad you are on the way up.

> May I suggest you raise the topic on the contributors mailing list,
> which is a better forum than github for such a broad discussion?
>

Yes, I should do that instead of trolling a random PR. Let me see if I
can find the time and the words.

osvalr:
:+1:

PR #49, opened 2016-07-13
[IMP] javascript-lint: Enable jshint configuration file

moylop260:
- [x] Enable configuration file for jshint to enable or disable checks and other custom configurations.
- [x] <s>Enable `console*` check</s> To detect a `console.log` we need all globals variables to use https://github.com/jshint/jshint/issues/2979#issuecomment-232212913 ... to add all variables we need all source js files... then this is not static check... search other way
- [x] Disable https://jslinterrors.com/expected-an-assignment-or-function-call

### DIFF
--- a/pylint_odoo/checkers/modules_odoo.py
+++ b/pylint_odoo/checkers/modules_odoo.py
@@ -83,6 +83,10 @@ ODOO_MSGS = {
 DFTL_README_TMPL_URL = 'https://github.com/OCA/maintainer-tools' + \
     '/blob/master/template/module/README.rst'
 DFTL_EXTFILES_TO_LINT = ['xml', 'csv', 'po', 'js', 'mako']
+DFTL_JSHINTRC = os.path.join(
+    os.path.dirname(os.path.dirname(os.path.realpath(__file__))),
+    'examples', '.jshintrc'
+)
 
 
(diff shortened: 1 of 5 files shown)


dreispt:
LGTM :+1:

pedrobaeza:
Please rebase

pedrobaeza:
Rebase for allowing to merge

yvaucher:
LGTM

Can we exclude also a directory with this config file ?
For example: https://travis-ci.org/OCA/geospatial/jobs/170084640

I have some js lib added with sphynx in <module_name>/doc/_static how to avoid the check on those?

moylop260:
Sorry for delay.


Using jshint is not posible use a easiest way to detect `console` calls.
Check my last comment about: https://github.com/jshint/jshint/issues/2979#issuecomment-232479489

I tested other tool called [eslint](http://eslint.org/docs/rules/no-console) and there is a explicit check to avoid use console easiest.

The bad news is that eslint requires a upgrade of nvm to v6 (travis use a older version by default).
This step is posible and valid, but could create a slow build for lints. (1.5 minutes to 3.5 minutes)

IMHO eslint is a excellent tool for lint of js files if you are agree I'll create other PR with this tool.

PR #51, opened 2016-07-27
[ADD] consider-merging-classes-inherited

moylop260:
If you have duplicated inherit:
```python
...
 class TestModel(models.Model):
     _inherit = 'res.company'`
...

...
 class TestModel2(models.Model):
     _inherit = 'res.company'`
...
```

Then the message `module1/__init__.py:1: [R7980(consider-merging-classes-inherited), ] Consider merging classes inherited to res.company.`

- [ ] Support `_name .. _inherit` valid case
- [ ] Add test case with `_inherit = []`

### DIFF
--- a/pylint_odoo/checkers/modules_odoo.py
+++ b/pylint_odoo/checkers/modules_odoo.py
@@ -3,6 +3,8 @@
 
 import os
 
+import astroid
+
 from pylint.checkers import utils
 
 from .. import misc, settings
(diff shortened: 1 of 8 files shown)


pedrobaeza:
This case can happen nowadays with the duality of old/new API because you need to do some things in one class and other things in another class. Example: https://github.com/OCA/product-attribute/blob/8.0/product_multi_image/models/product_product.py

moylop260:
@pedrobaeza 
Thank for feedback

This check just support basic inherit `_inherit='model.name'`
I see a `_inherit = [var, 'model.name']` then this check won't emit a message here.
I will add new test case with inherit list.

I'll change this check for valid cases if a `_name` is used then skip it, because is a valid case.

moylop260:
@pedrobaeza 
Fixed for cases where a `_name = 'valid.inherit'` or `_inherit=['valid.inherit1', 'valid.inherit2']` are used.
Thanks again.

pedrobaeza:
Thanks :+1:

dreispt:
:+1:

PR #52, opened 2016-07-27
[ADD] old-api7-method-defined: Emit message if the method defined have (self, cr, uid)

moylop260:
```python
def my_method(self, cr, uid, ids, context=None):  # method old api v7
    pass
```

Idea for v9 projects to easiest migration to v10 where the old api should be deprecated.
NOTE: This PR don't enable the check in MQT, but is required for our local process, we can discuss the enabling in MQT project.

### DIFF
--- a/pylint_odoo/checkers/no_modules.py
+++ b/pylint_odoo/checkers/no_modules.py
@@ -175,6 +175,11 @@ ODOO_MSGS = {
         'method-inverse',
         settings.DESC_DFLT
     ),
+    'R%d10' % settings.BASE_NOMODULE_ID: (
+        'Method defined with old api version 7',
+        'old-api7-method-defined',
+        settings.DESC_DFLT
+    ),
 }
 
 DFTL_MANIFEST_REQUIRED_KEYS = ['license']
@@ -370,7 +375,8 @@ class NoModuleChecker(BaseChecker):
 
     @utils.check_messages('api-one-multi-together',
                           'copy-wo-api-one', 'api-one-deprecated',
-                          'method-required-super')
+                          'method-required-super', 'old-api7-method-defined',
+                          )
     def visit_functiondef(self, node):
         """Check that `api.one` and `api.multi` decorators not exists together
         Check that method `copy` exists `api.one` decorator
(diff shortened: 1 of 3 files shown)


moylop260:
FYI It was used to help odoo
https://twitter.com/fpodoo/status/763506369228001280

pylint_odoo/checkers/no_modules.py
```diff
@@ -408,6 +414,11 @@ def visit_functiondef(self, node):
                 self.add_message('method-required-super',
                                  node=node, args=(node.name))
 
+        if self.linter.is_message_enabled('old-api7-method-defined'):
+            first_args = [arg.name for arg in node.args.args][:3]
+            if first_args == ['self', 'cr', 'uid']:
```
pedrobaeza: You should just check if `cr` and `uid` are in args I think.

moylop260: Thanks for feedback.
Do you have a example to change it?

FYI we are using Antony Lesuisse and Fabien Pinckaers' regex
https://twitter.com/fpodoo/status/763480940827938816

They are using `self`: `def .*self.*cr.*uid`

pedrobaeza: I'm thinking about other methods that passes these 2 arguments are for sure in old API. Methods outside a class don't have `self`. You should also look for `cursor` and `user`, that are other aliases. And finally, I think we can check also if there are instructions like `self.pool`.

moylop260: > Methods outside a class don't have `self`

Technically is not a `method` It is a `function`
The focus of this check are just `methods` (inside a class)

A good practice of pylint is create separated checks for methods and functions
We could create a new one with `old-api7-function-defined`
NOTE: I saw the hook and migration scripts using `cr` (without `uid`), but I don't saw functions using `cr, uid`, Do you have a example to create the new check?

> You should also look for `cursor` and `user`

Good point!

> I think we can check also if there are instructions like self.pool.

Good point too to detect use of old api, but Its is offtopic for this check, we could create `use-deprecated-attribute`

moylop260: Alias added
Thanks

pedrobaeza:
Please rebase

moylop260:
rebased
thanks

pedrobaeza:
:+1:

dreispt:
:+1:

alan196:
:+1:

PR #56, opened 2016-07-27
[ADD] odoo-addons-relative-import

moylop260:
Inspired from https://github.com/Vauxoo/stock-logistics-workflow/pull/16#discussion_r71438218

### DIFF
--- a/pylint_odoo/checkers/modules_odoo.py
+++ b/pylint_odoo/checkers/modules_odoo.py
@@ -4,7 +4,6 @@
 import os
 
 import astroid
-
 from pylint.checkers import utils
 
 from .. import misc, settings
--- a/pylint_odoo/checkers/modules_odoo.py
+++ b/pylint_odoo/checkers/modules_odoo.py
@@ -170,6 +170,8 @@ class ModuleChecker(misc.WrapperModuleChecker):
 
     def _get_odoo_module_imported(self, node):
         odoo_module = []
+        if not hasattr(astroid, 'ImportFrom'):
+            astroid.ImportFrom = astroid.From
         if isinstance(node, astroid.ImportFrom) and \
                 ('openerp.addons' in node.modname or
                  'odoo.addons' in node.modname):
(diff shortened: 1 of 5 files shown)


dreispt:
:+1:

max3903:
:+1:

pedrobaeza:
I don't get what is the problem you want to address.

moylop260:
This check catch the following case:

- `addons/my_module/tests/test_demo.py` with the following content:
 - `from openerp.addons.my_module import something`

But if you are `my_module` then you don't need import with `from openerp.addons.my_module...` because you are in we should import with `from .. my_module import something`

It's a variant of a relative-import but using the dynamic sys.modules of odoo.

pylint_odoo/checkers/modules_odoo.py
```diff
@@ -77,6 +78,12 @@
         'wrong-tabs-instead-of-spaces',
         settings.DESC_DFLT
     ),
+    'W%d50' % settings.BASE_OMODULE_ID: (
+        'Odoo addons relative import, should use "." '
```
pedrobaeza: A more explanatory text should be: "Same Odoo module relative import. You should use..."

pedrobaeza: Sorry, it should be "Same Odoo module absolute import. You should use..."

moylop260: Thanks for feedback
Change applied!

pedrobaeza:
Can you check the Travis error?

dreispt:
@moylop260 AFAICT for controllers we can't use this relative import

moylop260:
@dreispt 
Thanks for feedback.
I don't understand you
Could you send us a example?

dreispt:
When you want to import from the inside the same module, relative imports are OK.
But what about when you want to import from other modules?
Odoo always uses an abolute path, and this is commonly used in controllers.
Example: https://github.com/odoo/odoo/blob/9.0/addons/website_sale/controllers/main.py#L10

moylop260:
That example is a good case and not is emitted a message there because: The module is `website_sale` and you are importing `website`

The wrong case is if the module is `website_sale` and you are importing `website_sale`

dreispt:
Ok then.
Note that CI is red.

moylop260:
FYI error unrelated
FIX https://github.com/OCA/pylint-odoo/pull/61

pedrobaeza:
Please rebase

moylop260:
rebased
Thanks @pedrobaeza

pedrobaeza:
:+1:

PR #57, opened 2016-07-28
[REF] import-error: Skip openerp or odoo import because is a runtime import

moylop260:
If you use a IDEs with `pylint` checkers supports then this change is for you.
Suppress the error `import-error` for `import openerp.addons.mymodule` cases because is a runtime `import`
Same case for `import openerp` isn't a runtime import but many times we don't wish added it to our `PYTHONPATH` or install it.

If you use `PYTHONPATH` then the import-error isn't emitted and this change won't affect you

### DIFF
--- a/pylint_odoo/augmentations/main.py
+++ b/pylint_odoo/augmentations/main.py
@@ -31,6 +31,18 @@ def is_valid_openerp_osv_deprecated(node):
     return False
 
 
+def is_openerp_import(node):
+    """Verify if the node is a import openerp
+    :returns: True if is a openerp import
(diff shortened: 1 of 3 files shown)


pedrobaeza:
I really prefer to keep the import error and configure my environment for importing the corresponding path. This way, I also have auto-completion and other cool features from Python IDEs like PyCharm.

moylop260:
Yes that is right too.
This change won't affect you.

moylop260:
Closed because It is not longer commented and the unique comment is a disapproval

PR #61, opened 2016-08-13
[REF] tox: Freeze stable sha pylint20

moylop260:
We are using pylint==1.4 from requirements.txt
But we are using a unreleased pylint==2.0 (master branch) from tox in order to verify future compatibility with our checks.
Just now there is a error to install master package, then we need freeze a stable sha in order to avoid flaky errors in the future.


pedrobaeza:
:+1:  Fast-tracking to have green builds.

Issue #47, opened 2016-07-11
"Missing newline at end of file" check unwanted in some cases


yvaucher:
Hello,

Sometime we have sample files to test import. Those file are integrated in the module and saved as they were generated. We don't want to alter them at all as they are reference file for test purpose.

Thus we should disable this check:

    l10n_ch_account_statement_base_import/init.py:1: [W7908(missing-newline-extrafiles), ] data/UBS_export.csv Missing newline

Or provide a folder in module structure that isn't checked as it isn't odoo data but data to be processed by the module.

Or maybe we only want to check the files that are listed in the `__openerp__.py`

While we are searching for a solution I recommend to disable this check.

Example: https://travis-ci.org/OCA/l10n-switzerland/jobs/143363706

moylop260:
@OCA/core-maintainers What do you think about?

dreispt:
Would having the check only on py and xml files do?

gurneyalex:
I honestly don't care about the presence or absence of a new line at the end of any file in an Odoo module. Enforcing it brings nothing to the readability of the code (or lack thereof), and is only a hassle to contributors. 

I'm *not* in favor of having a pylint check mandating this -> I'm :+1: on this issue.

moylop260:
For record, the reason of the check:

- The reviewers request add a newline at end of file. (You can find them with [search newline comments from oca](https://github.com/pulls?utf8=%E2%9C%93&q=is%3Apr+user%3AOCA+newline+))
 - https://github.com/OCA/web/pull/389#discussion_r71902501
 - https://github.com/OCA/social/pull/74#discussion_r68369033
 - https://github.com/OCA/l10n-italy/pull/208#discussion_r56490900
 - https://github.com/OCA/l10n-italy/pull/138#discussion_r47076461
 - https://github.com/OCA/partner-contact/pull/201#discussion_r44780362
 - https://github.com/OCA/l10n-switzerland/pull/221#issuecomment-196701210
 - https://github.com/OCA/vertical-medical/pull/39#discussion_r40938313

- The history: https://robots.thoughtbot.com/no-newline-at-end-of-file

- Github show:
![without newline](https://cloud.githubusercontent.com/assets/6644187/14660615/8f41a4ee-066c-11e6-9eb9-6f9e9e56f7b2.png)

- If you have a editor with a configuration to auto-add a newline then you will see a dummy change of newline in all files edited.
[...]

PR #63, opened 2016-08-20
[REF] checkers: Skip example folder to check lints

moylop260:
Fix https://github.com/OCA/pylint-odoo/issues/47

### DIFF
--- a/pylint_odoo/checkers/modules_odoo.py
+++ b/pylint_odoo/checkers/modules_odoo.py
@@ -438,8 +438,6 @@ class ModuleChecker(misc.WrapperModuleChecker):
         """
         self.msg_args = []
         for js_file_rel in self.filter_files_ext('js', relpath=True):
-            if 'lib' in os.path.dirname(js_file_rel).split(os.sep):
-                continue
             js_file = os.path.join(self.module_path, js_file_rel)
             errors = self.check_js_lint(js_file)
             for error in errors:
@@ -475,9 +473,6 @@ class ModuleChecker(misc.WrapperModuleChecker):
         self.msg_args = []
         for type_file in self.config.extfiles_to_lint:
             for ext_file_rel in self.filter_files_ext(type_file, relpath=True):
-                if 'lib' in os.path.dirname(ext_file_rel).split(os.sep):
-                    continue
-
                 ext_file = os.path.join(self.module_path, ext_file_rel)
                 countline = 0
                 with open(ext_file, 'rb') as fp:
@@ -498,8 +493,6 @@ class ModuleChecker(misc.WrapperModuleChecker):
         self.msg_args = []
         for type_file in self.config.extfiles_to_lint:
             for ext_file_rel in self.filter_files_ext(type_file, relpath=True):
-                if 'lib' in os.path.dirname(ext_file_rel).split(os.sep):
-                    continue
                 ext_file = os.path.join(self.module_path, ext_file_rel)
                 last_line = ''
                 with open(ext_file, 'rb') as fp:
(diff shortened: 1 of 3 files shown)


pylint_odoo/checkers/modules_odoo.py
```diff
@@ -438,8 +438,6 @@ def _check_javascript_lint(self):
         """
         self.msg_args = []
         for js_file_rel in self.filter_files_ext('js', relpath=True):
-            if 'lib' in os.path.dirname(js_file_rel).split(os.sep):
```
pedrobaeza: Why are you removing this?

moylop260: Now is included in `def filter_files_ext` method.

pedrobaeza: OK, got it

pylint_odoo/misc.py
```diff
@@ -145,19 +145,27 @@ def _check_{NAME_KEY}(self, module_path)
...
+        dirnames_to_skip = ['example', 'examples', 'sample', 'samples', 'lib']
         if not fext.startswith('.'):
             fext = '.' + fext
         fext = fext.lower()
         fnames = self.ext_files.get(fext, [])
+        for fname in fnames:
+            dirnames = os.path.dirname(fname).split(os.sep)
+            for dirname_to_skip in dirnames_to_skip:
+                if dirname_to_skip in dirnames:
+                    fnames.remove(fname)
```
pedrobaeza: You're removing the element of a list iterating over that list. You need to make a copy of the list: `for fname in list(fnames):`

moylop260: That is right to iterate `generator` objects but is not for not generator objects.
List is not `generator` object.

pedrobaeza: Please execute this code in a console:
```
lista = [1, 2, 3, 4, 5]
for s in lista:
    print s
    l.remove(s)
```
You will see this output:
```
1
3
5
```
So you can see that it's not correct.

moylop260: You can test the following script:

```python
print "Using list: all fine"
dicc = {'key1': 'value1', 'key2': 'value2'}
for key in dicc.keys():
    print "pop", dicc.pop(key)
print "dicc", dicc

print "Using iterator: RuntimeError: dictionary changed size during iteration"
dicc = {'key1': 'value1', 'key2': 'value2'}
for key in dicc.iterkeys():
    print "pop", dicc.pop(key)
print "dicc", dicc
```

moylop260: In your example, you need add the `break` sentence.

moylop260: I mean
```python
my_list = [1, 2, 3, 4, 5]
for item in my_list:
    print "item", item
    if item == 4:
        print "removed", item
        my_list.remove(item)
        break
print "my_list", my_list
```

You will get the following output:
```bash
item 1
item 2
item 3
item 4
removed 4
my_list [1, 2, 3, 5]
```

moylop260: @pedrobaeza 
I saw your point.
My point consider 1 remove without error.
But your point consider more than 1 remove with error.

You are right!
We could have more than 1 remove because we have 2 nested loops.

Thanks!

moylop260: fixed

pedrobaeza:
:+1:

StefanRijnhart:
Thanks :+1:

moylop260:
ping

yvaucher:
:+1:

Issue #59, opened 2016-08-02
Add "Must bump version" check


dreispt:
This is a suggestion for a new check:

When doing a PR, it should bump the module's version should be bumped, either if we are porting, adding feature, or fixing code.
The check should run only for PRs.

Special case: some PR might not involve modules (e.g: changir .travis.yml) - in these case the check should not fail.

max3903:
:+1:

moylop260:
Its is posible if you add `version` in required keys from configuration file of PR from mqt

https://github.com/OCA/maintainer-quality-tools/blob/master/travis/cfg/travis_run_pylint_pr.cfg#L14

moylop260:
Or are you talking about auto increase the version number to manifest file?

Like as a plugin to odoo modules of https://pypi.python.org/pypi/bumpversion/0.5.3?

sebastienbeau:
:+1: for adding the check.
bumpversion can be a good tool for incrementing the version

dreispt:
The point is not to autoincrement (that depends on the type of change made).
The point is to check that we don't forget to change the version number when doing a PR.
You don't even have to check if it increases, it's enough to check if it changed.

moylop260:
@dreispt 
Thanks for clarify
I got it

I'm agree with you

What about if 2 people are working in 2 different changes?
The PRs will tend to diverge with conflicts because change the same line.
If PR1 is merged the PR2 will show conflicts or viceversa.
And the reviewers will request to contributors that fix the conflict but this could be requested many times by each PR merged of the same module.

Generally in many projects the change of versions are controlled from stable branch to avoid this issues.

IMHO we should think use a workflow (auto or manual) to change it from stable branch.

elicoidal:
:+1: with the check
:-1: for auto-increment

dreispt:
@moylop260 I think that's a corner case we can live with. And if two people are changing the same module, odds are that you will have conflicts and need a rebase.

moylop260:
Could somebody help me to create this check?

If you want help use the following videos to learn
https://asciinema.org/a/26890
https://asciinema.org/a/26889

moylop260:
Maybe this check is not a `pylint-odoo` one.
Because we need extract a `git diff` and verify if there is a `manifest version` change and pylint-odoo is not using a `git diff`

[...]

PR #64, opened 2016-08-21
[REF] Add bumpversion utility support

moylop260:
Integrate [bumpversion](https://pypi.python.org/pypi/bumpversion) utility to auto increase the version `{major.minor.patch}` using the following commands:
  - For increase .patch
     -  `bumpversion patch --tag`
  - For increase .minor
     -  `bumpversion minor --tag`
  - For increase .major
     - `bumpversion major --tag`

To sign the commit for auto-deploy for travis
  - `git commit --amend -s`
  - `git push origin master --tags`

Allow get the version installed with:
`python -c "import pylint_odoo; print pylint_odoo.__version__"`

### DIFF
--- a/pylint_odoo/__init__.py
+++ b/pylint_odoo/__init__.py
@@ -1,4 +1,7 @@
 
+__version__ = "1.3.2"
+
+
 from . import checkers
 from . augmentations.main import apply_augmentations
 
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,0 +1,13 @@
+# To auto increase the version {major.minor.patch} use the commands:
+#   bumpversion (major|minor|patch)
+# To sign commit for travis auto-deploy:
+#   git commit --amend -s
+#   git push origin master --tags
+
+[bumpversion]
+current_version = 1.3.2
+commit = True
+tag = True
+
+[bumpversion:file:pylint_odoo/__init__.py]
+


pylint_odoo/__init__.py
```diff
@@ -1,4 +1,7 @@
 
+__version__ = "1.3.2"
```
moylop260: FYI for the position of this variable I'm using the pep: https://www.python.org/dev/peps/pep-0008/#module-level-dunder-names

dreispt:
:+1: !

pedrobaeza:
I don't see this very interesting for Odoo, but we need a pylint-odoo check that checks the change of version number on each PR.

moylop260:
I'm thinking how to create it, because this is a non-common check (we need review a module changed for a particular environment, in local we don't like show this check and maybe in our forks too)...

moylop260:
Well, we have 2 things here.

 - A check for each pr of odoo modules: https://github.com/OCA/pylint-odoo/issues/59
 - Use of [bumpversion](https://pypi.python.org/pypi/bumpversion) for easiest release of pylint-odoo

FYI this PR talk about the second one.

pedrobaeza:
OK, I understand, but I don't see very useful run a command (and learn it) instead of changing the version manually.

moylop260:
We can change it manually.

This utility help us to process of:
 - Change of version in files
 - git commit
 - git tag

It's so easy, but I even forge a step, However with bumpversion I just run: `bumpversion patch` and the 3 steps are ran automatically.

But I don't have problem.
We can with the manual process :(

PR #65, opened 2016-08-22
[ADD] prefer-other-formatting: Avoid use of ''.format

moylop260:
ASCII errors could be caused by use string ''.format
You could see the following cases:
`u'%s' % u'á'`  # without error
`'%s' % u'á'`  # without error

`u'{}'.format(u'á')`  # without error
`'{}'.format(u'á')`  # error
You will see the error:
`UnicodeEncodeError: 'ascii' codec can't encode character u'\xe1' in position 1: ordinal not in range(128)`

Similar case for
`'%(var)s' % {'var': u'á'}`  # without error
`'{var}'.format(var=u'á')`  # error

It's agree with our guideline https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#idioms
`Prefer % over .format(), prefer %(varname) instead of positional. This is better for translation and clarity.`

### DIFF
--- a/pylint_odoo/checkers/no_modules.py
+++ b/pylint_odoo/checkers/no_modules.py
@@ -180,6 +180,12 @@ ODOO_MSGS = {
         'old-api7-method-defined',
         settings.DESC_DFLT
     ),
+    'R%d11' % settings.BASE_NOMODULE_ID: (
+        'Prefer "%(varname)"s over .format(). This is better for translation '
+        ' and clarity.',
+        'prefer-other-formatting',
+        settings.DESC_DFLT
+    ),
 }
 
 DFTL_MANIFEST_REQUIRED_KEYS = ['license']
@@ -281,9 +287,11 @@ class NoModuleChecker(BaseChecker):
 
     @utils.check_messages('translation-field', 'invalid-commit',
                           'method-compute', 'method-search', 'method-inverse',
-                          'sql-injection',
+                          'sql-injection', 'prefer-other-formatting',
                           )
     def visit_call(self, node):
+        if self.get_func_name(node.func) == 'format':
+            self.add_message('prefer-other-formatting', node=node)
         if node.as_string().lower().startswith('fields.'):
             args = misc.join_node_args_kwargs(node)
             for argument in args:
--- a/pylint_odoo/checkers/no_modules.py
+++ b/pylint_odoo/checkers/no_modules.py
@@ -290,7 +290,9 @@ class NoModuleChecker(BaseChecker):
                           'sql-injection', 'prefer-other-formatting',
                           )
     def visit_call(self, node):
-        if self.get_func_name(node.func) == 'format':
+        node_infer = utils.safe_infer(node.func)
+        if utils.is_builtin_object(node_infer) and \
+                self.get_func_name(node.func) == 'format':
             self.add_message('prefer-other-formatting', node=node)
         if node.as_string().lower().startswith('fields.'):
             args = misc.join_node_args_kwargs(node)
--- a/pylint_odoo/test/main.py
+++ b/pylint_odoo/test/main.py
@@ -43,6 +43,7 @@ EXPECTED_ERRORS = {
     'odoo-addons-relative-import': 4,
     'old-api7-method-defined': 2,
     'openerp-exception-warning': 3,
+    'prefer-other-formatting': 4,
     'redundant-modulename-xml': 1,
     'rst-syntax-error': 2,
     'sql-injection': 6,
--- a/pylint_odoo/test_repo/pylint_deprecated_modules/openerp/__init__.py
+++ b/pylint_odoo/test_repo/pylint_deprecated_modules/openerp/__init__.py
@@ -1,3 +1,4 @@
 # -*- coding: utf-8 -*-
 
 from . import osv  # pylint: disable=W0402
+_ = str


moylop260:
FYI I ran this check on oca projects:
`pylint --msg-template="{path}:{line}" -rn --load-plugins=pylint_odoo -d all -e prefer-other-formatting oca-projects/*/*`

I saw the message emitted for file: [mrp_production_partner_note/models/procurement_order.py](https://github.com/OCA/manufacture/blob/5fde4e9a3cfde9f3cfcdacb71ad1619afdca63cf/mrp_production_partner_note/models/procurement_order.py#L19)

I have created a POC to reproduce the ascii error and the result was: https://github.com/OCA/manufacture/issues/146

sbidoul:
Why does OCA encourage `%` over `format()`? In the python 3 documentation, it's the other way around: https://docs.python.org/3//library/stdtypes.html#old-string-formatting

moylop260:
Really is not a OCA guideline... Its a ODOO guideline if you review [odoo-guidelines](https://www.odoo.com/documentation/8.0/reference/guidelines.html) and search `Idioms` section you will see: `Prefer % over .format(), prefer %(varname) instead of position (This is better for translation)`

Using `''.format` We will have ASCII errors. Example: https://github.com/OCA/manufacture/issues/146

In the example used from pep: https://docs.python.org/3//library/stdtypes.html#old-string-formatting is using `format_string % values` that is good and agree with our guideline.
The bad case is: `'format_string'.format(values)` you could reproduce the errors from main description.

sbidoul:
The part saying "this is better for translation" refers to positional arguments. With `.format` we can use `{varname}`. So that is not relevant, I guess.

Regarding the unicode decoding error, this does not happen when using `u'...'.format()`, right? So I'd prefer to have pylint odoo detect `'...'.format()` (with a str instead of unicode) than completely rejecting the use of format, which is useful and future-proof.

moylop260:
Detect a `u'...'.format()` is a good idea.
But the problem is that we can have:
`var = '...'`
`var.format()`
And we can't detect this cases because is a dynamic one.

Maybe a dynamic check could detect them or maybe we could follow the current guideline of prefer `%s` or `%(varname)s` without `.format` in order to avoid future ASCII errors and have a static checker.

> and future-proof..

The `%(var)s % values` exists from `C` year 1980, this way is more future-proof that `''.format`.
In fact, in version 3.0 we have a new formatting text, but we don't have a deprecated `%(var)s` or `''.format()`

But I have a open mind in order to fix the current issues that are stealthy errors using `.format`

moylop260:
FYI we have today other issue-fix-rebuild-feedback loop of this matter: https://github.com/OCA/server-tools/pull/544

lasley:
I'm with @sbidoul on this - the `u'...'.format()`is preferred for me, mainly for testing. I can much more easily mock the `.format` method on a string in order to obtain caller args vs. the `%` - which I actually have no idea how to test outside of final result.

moylop260:
I don't have clear what is the advantage of use `.format` vs `%(arg)s`.

I see the same easily and the same final result.

print "My var1 is: %(var1)s" % dict(var1='value1')
print "My var1 is: {var1}".format(var1='value1')

We can use `u'string'.format` to detect them but we can detect `variable_string.format`.

The format just add additional maintenance cruft. (issue-fix-rebuild-feedback loop)

Yajo:
The `%` form was even proposed for deprecation, although that did not happen.
`.format()` is more powerful indeed.
However, for non-translatable strings, it makes no difference (except for the unicode issue). You could check too if the module has a `from __future__ import unicode_strings`.

moylop260:
Thanks for reply

I need help me to understand.
I want to close this PR with a good justification of use .format and pay the cost living with current codification errors and issue-fix-feedback loop.

But I can't understand where we are using the power of format. (3 examples using power of format on OCA projects could be useful)

Yajo:
You can patch malformed origin strings more efficiently in other languages:

```
_("The %s %s") % ("red", "car")  # When translating into Spanish, "El rojo coche" has no sense
_("The {} {}").format("red", "car")  # In Spanish you could put "El {1} {0}" and it will have sense
```

... although I still think keyword args is a better solution, available for both.

Hoping some day Odoo fixes https://github.com/odoo/odoo/issues/8530, [we could benefit from `f""` strings](http://stackoverflow.com/a/36645589/1468388).

Just my 2 cents

moylop260:
- pylint can detect the following valid cases to skip the check:
 - Use of future unicode literals thanks to @Yajo 
  ```python
from __future__ import unicode_literals  # We don't need `u` prefix in format string to avoid unicode error
'string {}'.format('á')
```
 - `u` prefix in `u'string'.format`
  ```python
u'string {}'.format(u'á')
```

- pylint can't detect the following invalid cases:
 - String variable without `u` prefix
  ```python
string_variable = 'string {}'
string_variable.format(u'á')  # unicode error
```

IMHO the middle point is:
 - Allow `.format`
 - Force use of `from __future__ import unicode_literals` if a `.format` is detected in the code. (everybody wins)

Then I'll close this PR and I'll create other one with this behaviour (please, comment if you are not agree)

**Comment by Yajo** [comment]
It seems a good choice :wink:.

However using the future import could, in some corner cases, break other parts of the code. If it is so, I hope there's a way of disabling this check (`# noqa: something` or similar).

But for 99%, seems quite good option. :+1:

**Comment by Yajo** [comment]
Wow, corner case just hit my face:

```
  File "/usr/lib/python2.7/site-packages/werkzeug/serving.py", line 177, in run_wsgi
    execute(self.server.app)
  File "/usr/lib/python2.7/site-packages/werkzeug/serving.py", line 165, in execute
    application_iter = app(environ, start_response)
  File "/opt/odoo/available/OCA/OCB/openerp/service/server.py", line 291, in app
    return self.app(e, s)
  File "/opt/odoo/available/OCA/OCB/openerp/service/wsgi_server.py", line 216, in application
    return application_unproxied(environ, start_response)
  File "/opt/odoo/available/OCA/OCB/openerp/service/wsgi_server.py", line 202, in application_unproxied
    result = handler(environ, start_response)
  File "/opt/odoo/available/OCA/OCB/openerp/http.py", line 1298, in __call__
    return self.dispatch(environ, start_response)
  File "/opt/odoo/available/OCA/OCB/openerp/http.py", line 1272, in __call__
    return self.app(environ, start_wrapped)
  File "/usr/lib/python2.7/site-packages/werkzeug/wsgi.py", line 579, in __call__
    return self.app(environ, start_response)
  File "/opt/odoo/available/OCA/OCB/openerp/http.py", line 1436, in dispatch
    ir_http = request.registry['ir.http']
  File "/opt/odoo/available/OCA/OCB/openerp/http.py", line 353, in registry
    return openerp.modules.registry.RegistryManager.get(self.db) if self.db else None
  File "/opt/odoo/available/OCA/OCB/openerp/modules/registry.py", line 339, in get
    update_module)
  File "/opt/odoo/available/OCA/OCB/openerp/modules/registry.py", line 370, in new
    openerp.modules.load_modules(registry._db, force_demo, status, update_module)
  File "/opt/odoo/available/OCA/OCB/openerp/modules/loading.py", line 350, in load_modules
    force, status, report, loaded_modules, update_module)
  File "/opt/odoo/available/OCA/OCB/openerp/modules/loading.py", line 255, in load_marked_modules
    loaded, processed = load_module_graph(cr, graph, progressdict, report=report, skip_modules=loaded_modules, perform_checks=perform_checks)
  File "/opt/odoo/available/OCA/OCB/openerp/modules/loading.py", line 152, in load_module_graph
    models = registry.load(cr, package)
  File "/opt/odoo/available/OCA/OCB/openerp/modules/registry.py", line 163, in load
    model = cls._build_model(self, cr)
  File "/opt/odoo/available/OCA/OCB/openerp/models.py", line 640, in _build_model
    '_original_module': original_module,
TypeError: type() argument 1 must be string, not unicode
```

Seems like unicodes cannot be used in `_name` or `_inherit`...

moylop260:
@Yajo 
Thanks for feedback
Yes, we have a problem using `from __future__ import unicode_literals` and `type()` method 😞
More info: http://stackoverflow.com/questions/19618031/unicode-literals-and-type

Then we are in the initial state again 😫 

| Case | style dynamic string  | Error? | 
| ---- | ------------- | ------------- |
| 1 | u"{var}".format(var=value)  | 🍏 No error |
| 2 | "{var}".format(var=value)  | 🍎 Unicode error    |
| 3 | my_str.format(var=value)  | 🍋 Maybe error <br/> (We don't know if `my_string` is unicode) |
| 4 |  u"%(var)s" % dict(var=value) | 🍏 No error  |
| 5 |  "%(var)s" % dict(var=value) | 🍏 No error  |
| 6 |  my_str % dict(var=value) | 🍏 No error  |

I like all green cases with `%(var)s` but many people prefer `.format`  and is good, but I prefer avoid the famous loop to fix unicode error for the case 2 and posible case 3.

The real problem for lint is emit a false error in case 3

Maybe, we could create the lint with:
if is case # 2 emit a error message.
if is case # 3 emit a warning message. But if is a "good" case we should use a disable comment or avoid use a variable `my_str` in a `format` and they are bad ideas too.

Yajo:
Your last conclusion seems most sensible.

PR #15, opened 2015-12-01
[FIX] OCA/MQT#302: Use of astroid version 1.3.8

moylop260:
fix https://github.com/OCA/maintainer-quality-tools/issues/302

PR #69, opened 2016-09-27
[REF] requirements: Update version of pylint and astroid

moylop260:
Update version of dependencies:
 - astroid==1.4.8
 - pylint==1.6.4
 - pylint-plugin-utils==0.2.4

For test use last sha:
 - https://github.com/PyCQA/astroid/commit/9979615
 - https://github.com/PyCQA/pylint/commit/05cc9a7

Useful in order to discuss new checks of pylint 1.6.4 from [MQT](https://github.com/oca/maintainer-quality-tools)

Fix https://github.com/OCA/pylint-odoo/pull/70

### DIFF
--- a/pylint_odoo/checkers/modules_odoo.py
+++ b/pylint_odoo/checkers/modules_odoo.py
@@ -191,9 +191,6 @@ class ModuleChecker(misc.WrapperModuleChecker):
                 'consider-merging-classes-inherited', node.lineno):
             return
         node_left = node.targets[0]
-        if not hasattr(astroid, 'ClassDef'):
-            # Compatibility with old pylint versions
-            astroid.ClassDef = astroid.Class
         if not isinstance(node_left, astroid.node_classes.AssName) or \
                 node_left.name not in ('_inherit', '_name') or \
                 not isinstance(node.value, astroid.node_classes.Const) or \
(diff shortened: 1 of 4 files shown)


pedrobaeza:
Why a freeze in an specific version?

moylop260:
In order to avoid issues like as https://github.com/OCA/maintainer-quality-tools/issues/302#issuecomment-161017215

pedrobaeza:
But that's something that is not going to happen anymore, and if so, then we act again, but this way we need to update each time there's a new version. API breakages are not common in the libraries.

moylop260:
> that is not going to happen anymore

Yes, because we fixed freezing the libraries from https://github.com/OCA/pylint-odoo/pull/15
In fact, new version of pylint and astroid is not compatible with old version (before this PR) of pylint-odoo

This repository is critical because is used from all OCA's repositories.
A fail here represent a fail in all OCA's PR and requires a fast-tracking upgrade in order to fix incompatibility (But we don't have many people creating PRs)

IMHO is better a stable critical repository with updates each time there is a new version that a bugly critical repository with auto-updates.

What do you think?

pedrobaeza:
OK, if the API compatibility is not assured, let's move this way.

moylop260:
Thanks

oscarolar:
:+1:

moylop260:
ping

agb80:
:+1: thansk for finish this @moylop260

Issue #74, opened 2016-10-12
Import in test files must not be protected


lmignon:
If a test file is run, the required libs must be installed. Therefore we must not protect the import since these files are loaded by odoo only when running tests

https://travis-ci.org/acsone/reporting-engine/jobs/167070681#L293

moylop260:
Thanks @lmignon 
Could you add the explicit rule for tests in https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#external-dependencies?

lmignon:
@moylop260 here it's https://github.com/OCA/maintainer-tools/pull/230

pedrobaeza:
Closing as there's a PR

PR #75, opened 2016-10-12
[REF] missing-import-error: Skip test file

moylop260:
since these files are loaded only when running tests and in such a case your module and their external dependencies are installed.
Fix https://github.com/OCA/pylint-odoo/issues/74

### DIFF
--- a/pylint_odoo/checkers/modules_odoo.py
+++ b/pylint_odoo/checkers/modules_odoo.py
@@ -293,8 +293,15 @@ class ModuleChecker(misc.WrapperModuleChecker):
         if import_category not in ('FIRSTPARTY', 'THIRDPARTY'):
             # skip if is not a external library or is a white list library
             return
-        self.add_message('missing-import-error', node=node,
-                         args=(module_name,))
+        relpath = os.path.relpath(
+            node.parent.file, os.path.dirname(self.manifest_file))
+        if os.path.dirname(relpath) != 'tests':
+            # missing-import-error rule doesn't apply to the test files
+            # since these files are loaded only when running tests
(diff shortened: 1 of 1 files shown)


jjscarafia:
:+1:

hbrunn:
I was hoping this also fixes https://travis-ci.org/OCA/web/jobs/167070120#L310, but this is another test, right?

hbrunn:
seems we'll also have to move https://github.com/OCA/pylint-odoo/pull/75/files#diff-4fed3e474b34a69c0631eb4c73cf6a56L299 ff into the scope of this test?

pedrobaeza:
No, it's this one, but we need to do 2 things:
- Label pylint-odoo with a new version.
- Change requirements.txt in MQT to use this new version.

pedrobaeza:
Yeah, but you're right about the other change that is needed.

moylop260:
> @pedrobaeza No, it's this one, but we need to do 2 things:
- Label pylint-odoo with a new version.
- Change requirements.txt in MQT to use this new version.

[We are installing from git directly](https://github.com/OCA/maintainer-quality-tools/blob/4cad6e57feaabfb805a9c7972ac8b8c53924d9b9/travis/travis_install_nightly#L12) then is not required a update of version of requirements.txt to work from MQT.

moylop260:
@hbrunn The case is different to https://github.com/OCA/pylint-odoo/issues/74
@lmignon proposed avoid a try-except in import of external libraries used from tests because we have the external dependency from manifest.

The other case have a external dependency used from test without define it in manifest.

PR #77, opened 2016-10-13
[IMP] manifest-version-format: Add valid_odoo_versions parameter

moylop260:
To force a valid version of odoo in the manifest version.

- [ ] After merge this one we need create a new PR for MQT to force the valid_odoo_versions as `VERSION` of the .travis.yml

Fix https://github.com/OCA/maintainer-quality-tools/issues/366


pedrobaeza:
But the idea is to check that for branch 10.0, the version number is 10.0.x.y.z, not to see if it's one of the valid versions (that for sure it will be: 8.0.x.y.z or 9.0.x.y.z).

moylop260:
Yes, that part will worked from MQT with `version` environment variable used from `.travis.yml` to set a unique valid version.

PR #82, opened 2016-10-15
[FIX] Whitelist `anybox.testing.openerp`

### LINKED ISSUE #81: Whitelist "anybox.testing.openerp" dependency
Requiring `anybox.testing.openerp` throws a [lint error](https://travis-ci.org/laslabs/web/jobs/167878323#L301). 

Pretty sure this is a known test dependency, so please whitelist. This should fix [current `web` 8.0 build](https://travis-ci.org/OCA/web/jobs/167564144) & OCA/web#445 Lint fail

lasley:
* Add `anybox.testing.openerp` - Fixes #81

### DIFF
--- a/pylint_odoo/checkers/modules_odoo.py
+++ b/pylint_odoo/checkers/modules_odoo.py
@@ -132,8 +132,8 @@ DFLT_IMPORT_NAME_WHITELIST = [
     # self-odoo
     'odoo', 'openerp',
     # Known external packages of odoo
-    'PIL', 'argparse', 'babel', 'dateutil',
-    'decorator', 'docutils', 'faces', 'feedparser',
+    'PIL', 'anybox.testing.openerp', 'argparse', 'babel',
+    'dateutil', 'decorator', 'docutils', 'faces', 'feedparser',
     'gdata', 'gevent', 'greenlet', 'jcconv', 'jinja2',
     'ldap', 'lxml', 'mako', 'markupsafe', 'mock', 'odf',
     'ofxparse', 'openid', 'passlib', 'pkg_resources',


lasley:
I can only comment on the `vertical-medical` ones, but it looks like it is being used to speed the tests up. I will likely remove this dependency during upgrades to favor `odoo.tests.common.SavepointCase`.

Maybe something that we did not historically have? That vertical-medical code is from like v6 or 7

moylop260:
I got it
Thanks for clarify

IMHO We should merge this PR in order to avoid superfluous changes in manifest file.
and after if is posible use `odoo.tests.common.SavepointCase` instead, we should add `anybox.testing.openerp` it to [deprecated-modules-list](https://github.com/OCA/maintainer-quality-tools/blob/a6e6aea1f527c74c2d905bb595d1216a791c4bb4/travis/cfg/travis_run_pylint_pr.cfg#L121) to see a beta message to invite to change it (But not a seudo-false red one)

Let's wait other reviewers to know point of view.

pedrobaeza:
But didn't we include an exception for not checking this on tests folder?

moylop260:
This matter is a little different.

We added a exception in order to avoid a `try-except:ImporError` in tests because the manifest have the dependency, then if the module is installed the tests will run fine.

But here we have a library that is not added in the manifest and is used in test, then if you don't have installed this library the module can be installed and the tests will be fail.

moylop260:
I mean

| Case | Is there library in manifest? | Is there library imported in test/? | Emit message of `missing-try-except-import` in test/ | Emit message of `missing-manifest` in test/ | 
|---|---|------------|------|-----|
| 1 | yes  | no    |   N/A  | N/A
| 2 |  yes | yes |     no thanks to https://github.com/OCA/pylint-odoo/pull/75 This case is https://github.com/OCA/pylint-odoo/issues/74  | N/A
| 3 |  no | yes |    no thanks to https://github.com/OCA/pylint-odoo/pull/75 | yes (This case is this PR)
| 4 |  no | no |     N/A | no

PR #85, opened 2016-11-03
[ADD] missing-return

moylop260:
If you use call a `super` method then you will need return the original
value.
If you want overwrite a original method then you need add documentation
of why and add a `pylint: disable=missing-return`


moylop260:
ping

pedrobaeza:
I think we can merge this.

Issue #83, opened 2016-10-24
I can not have demo and data with the same xml id because the next error is displayed:  [W7902(duplicate-xml-record-id), ]  Duplicate xml record id 


yaniaular:
**Error**

I can not have demo and data record with a same xml id because the next error is displayed: 

``` python
data/res_company.xml:4: [W7902(duplicate-xml-record-id), ]  Duplicate xml record id "base.main_company.noupdate0" in demo/res_company.xml
```

**Steps to reproduce the error:**
1. I have two files: data/res_company.xml and demo/res_company.xml
2. The main company is edited in both files: 

**data/res_company.xml** 

``` xml
<record id="base.main_company" model="res.company">
     <field name="new_field_id" ref="module.reference_example_1"/>
</record>
```

**demo/res_company.xml**

``` xml
<record id="base.main_company" model="res.company">
     <field name="example_field">300</field>
</record>
```

With this scene, the next error is displayed:

``` python
data/res_company.xml:4: [W7902(duplicate-xml-record-id), ]  Duplicate xml record id "base.main_company.noupdate0" in demo/res_company.xml
```

**Conclusion**

 I think I can have data and demo with the same xml_id, it is valid.

I think this is a bug

dreispt:
Why would you need that?

guewen:
I think I had the same use case. You want to update or create data at the installation of the addon, but you want to configure a field with a particular value for the demo.

moylop260:
FYI fixed from https://github.com/OCA/pylint-odoo/issues/89 your feedback is welcome

PR #89, opened 2016-12-07
[REF] duplicate-xml-record-id: Get xml section from manifest to skip different origin

moylop260:
Fix OCA/pylint-odoo#83

Change test diff message in case of fails:
xml

### DIFF
--- a/pylint_odoo/checkers/modules_odoo.py
+++ b/pylint_odoo/checkers/modules_odoo.py
@@ -407,7 +407,7 @@ class ModuleChecker(misc.WrapperModuleChecker):
         return all_records
 
     def _check_duplicate_xml_record_id(self):
-        """Check duplicated XML-IDs inside of the files of 
+        """Check duplicated XML-IDs inside of the files of
         each manifest-section treated them separately
         :return: False if exists errors and
                  add list of errors in self.msg_args
--- a/pylint_odoo/checkers/modules_odoo.py
+++ b/pylint_odoo/checkers/modules_odoo.py
@@ -394,8 +394,11 @@ class ModuleChecker(misc.WrapperModuleChecker):
         """
         all_records = {}
         for record in records:
-            record_id = record.attrib.get('id', '') + '.' + "noupdate" + \
-                record.getparent().attrib.get('noupdate', '0')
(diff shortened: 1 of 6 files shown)


pylint_odoo/test/main.py
```diff
@@ -118,8 +120,8 @@ def test_20_expected_errors(self):
         pylint_res = self.run_pylint(self.paths_modules)
         # Expected vs found errors
         real_errors = pylint_res.linter.stats['by_msg']
-        self.assertEqual(sorted(real_errors.items()),
-                         sorted(EXPECTED_ERRORS.items()))
+        self.assertEqual(json.loads(json.dumps(EXPECTED_ERRORS)),
```
pedrobaeza: Why do you need to use json?

pylint_odoo/checkers/modules_odoo.py
```diff
@@ -409,11 +412,16 @@ def _check_duplicate_xml_record_id(self):
                  add list of errors in self.msg_args
         """
         self.msg_args = []
-        all_xml_ids = []
-        for xml_file in self.filter_files_ext('xml', relpath=False):
-            all_xml_ids.extend(self.get_xml_records(xml_file))
+        xml_records = []
+        for fname, section in self._get_manifest_referenced_files().items():
```
pedrobaeza: Add some comment here to explain that this only gets XML files from data section

moylop260: The method `_get_manifest_referenced_files` gets all type of files (Not only xml)

pedrobaeza: OK, I get it. Then the comment should be: `Check duplicated XML-IDs inside of the files of each section treated them separately`

moylop260: Oh! Are you talking about `docstrint`, right?
I got it

yaniaular:
:+1:

moylop260:
@dreispt @guewen 
This fix the issue https://github.com/OCA/pylint-odoo/issues/83?

ysantiago:
Good suggestion :+1:

moylop260:
Could we merge it?

PR #91, opened 2016-12-17
[FIX] test: Never use global into test

moylop260:
- Now we have a `assertEqual(dict1, dict2)` then a sum of checkers is not required.
- `global VARIABLE` is not a good idea from unittest, because could affect the result of other tests.
- Convert comments to docstring
- Clean code

### DIFF
--- a/pylint_odoo/misc.py
+++ b/pylint_odoo/misc.py
@@ -26,16 +26,6 @@ def get_plugin_msgs(pylint_run_res):
     return all_plugin_msgs
 
 
-def get_sum_fails(pylint_stats):
-    """Get a sum of all fails.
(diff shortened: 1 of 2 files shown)


lasley:
Oh neat, if I'm reading this code right it looks like this helps in OCA/maintainer-quality-tools#352, which I seem to have forgotten about.

moylop260:
Sorry I didn't found relation between this PR and https://github.com/OCA/maintainer-quality-tools/pull/352 could you help me to understand the part where are related?

moylop260:
You feedback is welcome @alan196 @jjscarafia @dreispt @pedrobaeza @oscarolar

Issue #76, opened 2016-10-13
duplicate-xml-fields: False red using 2 tree sub-views


moylop260:
Comment extracted from https://github.com/OCA/maintainer-quality-tools/pull/284#issuecomment-253470112 by @lasley 

> Regarding W7907(duplicate-xml-fields) - we ran into a false positive earlier, so I'm not so sure it would be good to fail builds on it. I'm trying to figure out what's causing it, but not so sure. Here's the info:
> - clouder-community/clouder#134
> - [Failure in Travis](https://travis-ci.org/clouder-community/clouder/jobs/167125877#L325)
> - [The code](https://github.com/clouder-community/clouder/blob/0.9.0/clouder/view.xml#L824)
> 
> The duplicate XML is being triggered for the `name` fields in *2M relation.

PR #96, opened 2016-12-23
[FIX] duplicate-xml-fields: False red using 2 tree sub-views *2M fields

moylop260:
Fix https://github.com/OCA/pylint-odoo/issues/76

### DIFF
--- a/pylint_odoo/checkers/modules_odoo.py
+++ b/pylint_odoo/checkers/modules_odoo.py
@@ -3,6 +3,7 @@
 
 import os
 import re
+import collections
 
 import astroid
 import isort
@@ -484,6 +485,16 @@ class ModuleChecker(misc.WrapperModuleChecker):
             all_fields.setdefault(field_xml, []).append(field)
         # Remove all keys which not duplicated
         for key, items in all_fields.items():
+            parents = [e.getparent() for e in items]
+            same_parent = [
+                item for item, count in collections.Counter(parents).items()
+                if count > 1]
+            # Remove elements which are in different parent element
+            # Parents in items but not in same_parent list
+            exclude = list(set(parents) - set(same_parent))
+            for item in items:
+                if item.getparent() in exclude:
+                    items.remove(item)
             if len(items) < 2:
                 all_fields.pop(key)
         return all_fields
--- a/pylint_odoo/checkers/modules_odoo.py
+++ b/pylint_odoo/checkers/modules_odoo.py
@@ -483,11 +483,11 @@ class ModuleChecker(misc.WrapperModuleChecker):
                 continue
             all_fields.setdefault(
                 (field_xml, field.getparent()), []).append(field)
-        # Remove all keys which not duplicated
-        single_fields = dict(((field_xml_name, parent_node), nodes) for
-                             (field_xml_name, parent_node), nodes in
-                             all_fields.items() if len(nodes) >= 2)
-        return single_fields
+        # Remove all keys which not duplicated by excluding them from the
+        # returning dict
+        return dict(((field_xml_name, parent_node), nodes) for
+                    (field_xml_name, parent_node), nodes in
+                    all_fields.items() if len(nodes) >= 2)
 
     def _check_duplicate_xml_fields(self):
         """Check duplicate field in all record of xml files of a odoo module.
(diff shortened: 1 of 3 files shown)


pylint_odoo/checkers/modules_odoo.py
```diff
@@ -484,6 +485,16 @@ def _get_duplicate_xml_fields(self, fields):
...
+            parents = [e.getparent() for e in items]
+            same_parent = [
+                item for item, count in collections.Counter(parents).items()
+                if count > 1]
+            # Remove elements which are in different parent element
+            # Parents in items but not in same_parent list
+            exclude = list(set(parents) - set(same_parent))
+            for item in items:
+                if item.getparent() in exclude:
+                    items.remove(item)
```
moylop260: @pedrobaeza 
Here we will have a issue with this remove, right?
cc @hugho-ad

moylop260: @hugho-ad 
Please, review the following good comments from @pedrobaeza and fix them
https://github.com/OCA/pylint-odoo/pull/63#r75580525

We have the same case here

pylint_odoo/test_repo/broken_module/model_view_odoo2.xml
```diff
@@ -85,5 +85,25 @@
...
+                    </field>
+                    <field name="value_ids">
+                        <tree>
+                            <field name="name"/>
+                        </tree>
+                    </field>
+                </form>
+            </field>
+        </record>
+
```
moylop260: @hugho-ad 
Could you add the following test case?
```xml
+        <!-- Double duplicate record field "name" in *2M field-->
 +        <record id="view_model_form5" model="ir.ui.view">
 +            <field name="name">view.model.form</field>
 +            <field name="model">test.model</field>
 +            <field name="arch" type="xml">
 +                <form string="Test model6">
 +                    <field name="picking_ids">
 +                        <tree>
 +                            <field name="name"/>
 +                            <field name="name"/>
 +                        </tree>
 +                    </field>
 +                    <field name="value_ids">
 +                        <tree>
 +                            <field name="name"/>
 +                            <field name="name"/> 
 +                        </tree>
 +                    </field>
 +                </form>
 +            </field>
 +        </record>
```

Here we should have 2 new messages emitted (increase the expected checkers)

moylop260: @hugho-ad 
FYI I ran the test and duplicated-xml-fields was emitted 7 times.
Maybe we need add the parent in the key in order to group all items, I mean:
```diff
- all_fields.setdefault(field_xml, []).append(field)
+ all_fields.setdefault((field_xml, field.getparent()), []).append(field)
```

And process even this multi key in all  other part of code where is used.

pylint_odoo/checkers/modules_odoo.py
pedrobaeza: I see that you have fixed it, so no problem now :wink:

lasley:
cc @YannickB - this will fix the duplicate xml ID lint in Clouder 😄

hugho-ad:
@moylop260 increased the expected checkers, and travis is green
please give your feadback

hugho-ad:
@moylop260 please give your feadback here

Regards

pylint_odoo/checkers/modules_odoo.py
```diff
@@ -469,23 +469,32 @@ def _check_redundant_modulename_xml(self):
...
+        all_fields = {}
+        for field in fields:
+            field_xml = field.attrib.get('name')
+            if not field_xml:
+                continue
+            parent = field.getparent()
+            parents.setdefault(
+                parent, {}).setdefault((field_xml, parent), []).append(field)
+            all_fields.setdefault((field_xml, parent), []).append(field)
+        return parents, all_fields
```
moylop260: You don't need 2 dictionaries to detect duplicated.
I mean, you could use just 1 dictionary since that you have the complex key:
 - field_xml <- str
 - parent <- node object (this is different although name string of parent is equal)

And the appended values are fields
Then you could detect them with
`[(field_xml, nodes) for (field_xml, parent), nodes in all_fields.items() if len(nodes) >= 2)`

hugho-ad:
@moylop260 comments Done! thks!

moylop260:
Feedback is welcome team

moylop260:
ping

yaniaular:
:+1:

PR #98, opened 2016-12-26
[FIX] method-NAME: Fix if value is not a string

moylop260:
Similar to https://github.com/OCA/pylint-odoo/pull/90
But for any type of value different to `string`


tomeyro:
:+1:

moylop260:
3 approvals but I don't want merge my own PR.
Could something merge it?

pedrobaeza:
Well, personally I think there's no problem for that, as it's totally justified, but don't worry, I merge.

PR #99, opened 2016-12-27
[ADD] renamed-field-parameter: detect deprecated field values (digits_compute, select) (#91)

tomeyro:
Add check to detect deprecated/renamed parameters in a field constructor.

Checking the parameters specified in: https://github.com/odoo/odoo/blob/10.0/odoo/fields.py#L29

image

### DIFF
--- a/pylint_odoo/checkers/no_modules.py
+++ b/pylint_odoo/checkers/no_modules.py
@@ -186,6 +186,11 @@ ODOO_MSGS = {
         'old-api7-method-defined',
         settings.DESC_DFLT
     ),
+    'W%d11' % settings.BASE_NOMODULE_ID: (
+        'Field parameter "%s" is no longer supported. Use "%s" instead.',
+        'renamed-field-parameter',
+        settings.DESC_DFLT
+    ),
 }
 
 DFTL_MANIFEST_REQUIRED_KEYS = ['license']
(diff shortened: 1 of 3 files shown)


pylint_odoo/checkers/no_modules.py
```diff
@@ -324,6 +369,15 @@ def visit_call(self, node):
                                 '_' + argument.arg + '_'):
                         self.add_message('method-' + argument.arg,
                                          node=argument_aux)
+                    elif (argument.arg in
+                            self.config.deprecated_field_parameters):
```
lasley: Non blocking, but you could remove a few lines by using an intermediary variable:

```
deprecated = self.config.deprecated_field_parameters
```

pylint_odoo/checkers/no_modules.py
```diff
@@ -306,9 +328,32 @@ class NoModuleChecker(BaseChecker):
...
+        :type colon_list: list
+        :returns: A dictionary with the values assigned to corresponding keys.
+        :rtype: dict
+
+        :Example:
+
+        >>> self.colon_list_to_dict(['colon:list', 'empty_key:'])
+        {'colon': 'list', 'empty_key': ''}
+        """
+        return dict([item.split(":") for item in colon_list])
```
lasley: No need to create a list:
```
>>> dict(item.split(':') for item in colon_list)
{'key2': 'val2', 'key1': 'val1'}
```

tomeyro:
Thanks for the comments @lasley. Updated code as suggested.

moylop260:
@JesusZapata 
Could you fix conflicts please?

JesusZapata:
@moylop260 
Done! I fixed the conflict!

PR #105, opened 2017-01-14
[IMP] Check attribute without translation parameter (OCA#104)

JesusZapata:
Check attribute without translation parameter
Referring to OCA#104


JesusZapata:
@moylop260 
What do you think about this PR?

PR #107, opened 2017-01-18
[ADD] except-pass: Emit message If a except:pass is used (OCA/#252)

JesusZapata:
The `pass` into block except is not a good practice!

  By including the `pass` we assume that our algorithm can continue to function after the exception occurred

  If you really need to use the `pass` consider logging that exception

  ```python
      try:
          sentences
      except:
          pass
         _logger.debug('Why the exception is safe....', exc_info=1))
  ```

fix https://github.com/OCA/maintainer-tools/pull/252

### DIFF
--- a/pylint_odoo/checkers/modules_odoo.py
+++ b/pylint_odoo/checkers/modules_odoo.py
@@ -117,6 +117,12 @@ ODOO_MSGS = {
         'missing-manifest-dependency',
         settings.DESC_DFLT
     ),
+    'W%d38' % settings.BASE_OMODULE_ID: (
+        'pass into block except. '
+        'If you really need to use the pass consider logging that exception',
+        'except-pass',
+        settings.DESC_DFLT
+    ),
     'W%d37' % settings.BASE_OMODULE_ID: (
         '%s The xml attribute is missing the translation="off" tag %s',
         'xml-attribute-translatable',
@@ -350,6 +356,15 @@ class ModuleChecker(misc.WrapperModuleChecker):
         for name, _ in node.names:
             if isinstance(node.scope(), astroid.Module):
                 self._check_imported_packages(node, name)
+
+    @utils.check_messages('except-pass')
+    def visit_tryexcept(self, node):
+        """Visit block try except"""
+        for handler in node.handlers:
+            if (not handler.name and
+                    len(handler.body) == 1 and
+                    isinstance(handler.body[0], astroid.node_classes.Pass)):
+                self.add_message('except-pass', node=handler)
 
     def _check_rst_syntax_error(self):
         """Check if rst file there is syntax error
--- a/pylint_odoo/test_repo/test_module/except_pass.py
+++ b/pylint_odoo/test_repo/test_module/except_pass.py
@@ -1,0 +1,62 @@
+# -*- coding: utf-8 -*-
+"""Test Except Pass usage"""
+
+
+class TestExceptPass(object):
+    """Test Except Pass class """
+
+    def test_method(self):
+        """Test method """
+        try:
+            raise Exception('Exception')
+        except Exception:  # except-pass
+            pass
+
+    def test_2_method(self):
+        """Test 2 method """
+        try:
+            raise Exception('Exception')
+        except Exception:
+            pass
+            print('Exception')
+
+    def test_3_method(self):
+        """Test 3 method """
+        try:
+            raise Exception('Exception')
+        except Exception as exception:
+            pass
+        if exception:
+            pass
+
+    def test_4_method(self):
+        try:
+            raise Exception('Exception')
+        except Exception, userError:
+            pass
+        if userError:
+            pass
+
+    def test_5_method(self):
+        try:
+            raise Exception('Exception')
+        except (Exception, IndexError) as exception:
+            pass
+        if exception:
+            pass
+
+    def test_6_method(self):
+        try:
+            raise Exception('Exception')
+        except (Exception, IndexError):  # except-pass
+            pass
+
+    def test_7_method(self):
+        try:
+            raise Exception('Exception')
+        except (Exception, IndexError, NameError), exception:
+            pass
+        except Exception:  # except-pass
+            pass
+        if exception:
+            pass
--- a/pylint_odoo/test_repo/test_module/except_pass.py
+++ b/pylint_odoo/test_repo/test_module/except_pass.py
@@ -6,14 +6,13 @@ class TestExceptPass(object):
     """Test Except Pass class """
 
     def test_method(self):
-        """Test method """
         try:
             raise Exception('Exception')
         except Exception:  # except-pass
             pass
 
     def test_2_method(self):
-        """Test 2 method """
+        """This pass is skip for body of except has more than one line """
         try:
             raise Exception('Exception')
         except Exception:
@@ -21,7 +20,7 @@ class TestExceptPass(object):
             print('Exception')
 
     def test_3_method(self):
-        """Test 3 method """
+        """This pass is skip for the exception is assigned"""
         try:
             raise Exception('Exception')
         except Exception as exception:
--- a/pylint_odoo/test/main.py
+++ b/pylint_odoo/test/main.py
@@ -58,6 +58,7 @@ EXPECTED_ERRORS = {
     'wrong-tabs-instead-of-spaces': 2,
     'eval-referenced': 5,
     'xml-syntax-error': 2,
+    'except-pass': 3,
     'attribute-string-redundant': 33,
     'renamed-field-parameter': 2,
     'xml-attribute-translatable': 1,


JesusZapata:
@moylop260 
What do you think about this PR?

pylint_odoo/checkers/modules_odoo.py
```diff
@@ -334,6 +340,14 @@ def visit_import(self, node):
             if isinstance(node.scope(), astroid.Module):
                 self._check_imported_packages(node, name)
 
+    @utils.check_messages('except-pass')
+    def visit_tryexcept(self, node):
+        """Visit block try except"""
+        for handler in node.handlers:
+            if handler.body:
```
moylop260: use `and` instead of nested `if` (in order to reduce mccabe rate)

JesusZapata: @moylop260 
Thanks!

pylint_odoo/checkers/modules_odoo.py
```diff
@@ -334,6 +340,14 @@ def visit_import(self, node):
             if isinstance(node.scope(), astroid.Module):
                 self._check_imported_packages(node, name)
 
+    @utils.check_messages('except-pass')
+    def visit_tryexcept(self, node):
+        """Visit block try except"""
+        for handler in node.handlers:
+            if handler.body and isinstance(handler.body[0],
```
moylop260: And `len(handler.body) == 1` because we need validate just that unique sentence.

```python
try:
  code
except:
   pass
   logger  # this case is valid, is other error but is valid for this one
```

JesusZapata: @moylop260 
Done!

pylint_odoo/checkers/modules_odoo.py
```diff
@@ -334,6 +340,14 @@ def visit_import(self, node):
             if isinstance(node.scope(), astroid.Module):
                 self._check_imported_packages(node, name)
 
+    @utils.check_messages('except-pass')
+    def visit_tryexcept(self, node):
+        """Visit block try except"""
+        for handler in node.handlers:
+            if len(handler.body) == 1 and \
```
moylop260: Avoid use `\` better use `(`
```python
  if (len(handler.body) == 1  and
      isinstace(...))
```

JesusZapata: @moylop260 
Done!
Thanks!

moylop260:
Could you fix conflicts?

JesusZapata:
@moylop260 
Done! I fixed this conflict!

moylop260:
@lasley @pedrobaeza @dreispt @lmignon 
your feedback is welcome

lasley:
Hrmmmm I'm torn on this one. Forgive my ignorance in pylint, but is it possible to check if the exception is used slightly later in code? I use a strategy similar to below fairly often:

```
err = None
for i in iter:
    try:
        i()
    except IException as err:
        pass
if err:
    raise err
```

moylop260:
That is a good point.
@JesusZapata 
Could you support that case, please?

lasley:
Yeah maybe we just leverage the pre-existing check to see whether a variable is used in the namespace? If the exception is assigned to a var, and the var is used, we can assume the exception is handled properly - or following the Zen of an explicit silence.

Thanks for taking this into consideration!

dreispt:
Than could be just a lint warning message, not breaking the build.

pylint_odoo/checkers/modules_odoo.py
```diff
@@ -351,6 +357,14 @@ def visit_import(self, node):
             if isinstance(node.scope(), astroid.Module):
                 self._check_imported_packages(node, name)
 
+    @utils.check_messages('except-pass')
+    def visit_tryexcept(self, node):
+        """Visit block try except"""
+        for handler in node.handlers:
```
moylop260: Here you could validate if the except has a ` as ` in order to skip the check because the new variable could be used.
It this variable is not used then we will have a `variable-unused`

This is the easiest way to skip the check using the @lasley recommendation

moylop260:
> @dreispt Than could be just a lint warning message, not breaking the build.

Yeah, from MQT we could enabled it like as a beta check

pylint_odoo/checkers/modules_odoo.py
JesusZapata: @moylop260 @lasley 
Done, Now we exclude if the exception is assigned to one variable

pylint_odoo/checkers/modules_odoo.py
```diff
@@ -351,6 +357,16 @@ def visit_import(self, node):
...
+            if (len(handler.body) == 1 and
+                    isinstance(handler.body[0], astroid.node_classes.Pass) and
+                    not isinstance(node.handlers[0].name,
```
moylop260: `and node.handlers and ` in order to avoid IndexError using `node.handlers[0]`

pylint_odoo/test_repo/test_module/except_pass.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+class TestExceptPass(object):
+    """Test Except Pass class """
+
+    def test_method(self):
+        """Test method """
+        try:
+            raise Exception('Exception')
+        except Exception:
```
moylop260: Add other test case when there aren't handlers.
```python
       try:
            raise Exception('Exception')
       except Exception:
```

and other ones using many handlers

```python
       try:
            raise Exception('Exception')
       except Exception, UserError:
```


```python
       try:
            raise Exception('Exception')
       except (Exception, UserError) as exception:

Comment (JesusZapata) [inline pylint_odoo/test_repo/test_module/except_pass.py]: @moylop260 
Done,
This cases work fine.

File: pylint_odoo/checkers/modules_odoo.py (hunk 5 above)
Comment (JesusZapata) [inline pylint_odoo/checkers/modules_odoo.py]: @moylop260 
Done, added another condition to if.

File: pylint_odoo/checkers/modules_odoo.py [hunk 7]
```diff
@@ -118,6 +118,12 @@
         settings.DESC_DFLT
     ),
     'W%d37' % settings.BASE_OMODULE_ID: (
+        'pass into block except. '
+        'If you really need to use the pass consider logging that exception',
+        'except-pass',
+        settings.DESC_DFLT
+    ),
+    'W%d38' % settings.BASE_OMODULE_ID: (
```
moylop260: The `except-pass` should be 38 not 37

JesusZapata:
@moylop260 
I have attached more cases and now check if there are several handlers.

pylint_odoo/test_repo/test_module/except_pass.py
```diff
@@ -0,0 +1,56 @@
...
+            pass
+        if exception:
+            pass
+
+    def test_6_method(self):
+        try:
+            raise Exception('Exception')
+        except (Exception, IndexError, NameError), exception:
+            pass
+        except Exception:
```
moylop260: Add a comment where the message should be emitted:
`except Exception:  # except-pass`

pylint_odoo/test_repo/test_module/except_pass.py
```diff
@@ -0,0 +1,56 @@
...
+            raise Exception('Exception')
+        except Exception as exception:
+            pass
+        if exception:
+            pass
+
+    def test_4_method(self):
+        try:
+            raise Exception('Exception')
+        except Exception, UserError:
```
moylop260: Use snake_name for `UserError`

pylint_odoo/test_repo/test_module/except_pass.py
```diff
@@ -0,0 +1,56 @@
...
+            pass
+        if UserError:
+            pass
+
+    def test_5_method(self):
+        try:
+            raise Exception('Exception')
+        except (Exception, IndexError) as exception:
```
moylop260: Add the case where there are many Exceptions but without alias `except (Exception, IndexError):`

pylint_odoo/checkers/modules_odoo.py
```diff
@@ -351,6 +357,15 @@ def visit_import(self, node):
             if isinstance(node.scope(), astroid.Module):
                 self._check_imported_packages(node, name)
 
+    @utils.check_messages('except-pass')
+    def visit_tryexcept(self, node):
+        """Visit block try except"""
+        for handler in node.handlers:
+            if (not handler.name and
+                    len(handler.body) and
```
moylop260: `s/len(handler.body)/handler.body/g`

pylint_odoo/test_repo/test_module/except_pass.py
```diff
@@ -0,0 +1,62 @@
...
+            raise Exception('Exception')  # except-pass
+        except Exception:
+            pass
+
+    def test_2_method(self):
+        """Test 2 method """
+        try:
+            raise Exception('Exception')
+        except Exception:
+            print('Exception')
```
moylop260: Use `pass` before of `print`

JesusZapata: @moylop260 
Done,
I have validated the case of multiple handler.body

moylop260:
@lasley 
Could you review the case requested, please?

moylop260:
@JesusZapata 
Could you squash all commits in one, please?

JesusZapata:
@moylop260 
I made rebase.

moylop260:
@lasley Do you have any comment about?

JesusZapata:
@lasley 
Only documented when skip the except-pass

PR #116, opened 2017-02-07
[FIX] fields: Validate a assign directly from root class

JesusZapata:
Avoid false messages if a variable `fields` is used from other side different to assign new columns for odoo classes

Referring to https://github.com/Vauxoo/pylint-odoo/issues/52

### DIFF
--- a/pylint_odoo/checkers/no_modules.py
+++ b/pylint_odoo/checkers/no_modules.py
@@ -375,7 +375,9 @@ class NoModuleChecker(BaseChecker):
                           'renamed-field-parameter'
                           )
     def visit_call(self, node):
-        if node.as_string().lower().startswith('fields.'):
+        if 'fields' == self.get_func_lib(node.func) and \
+                isinstance(node.parent, astroid.Assign) and \
+                isinstance(node.parent.parent, astroid.ClassDef):
             args = misc.join_node_args_kwargs(node)
             index = 0
             field_name = ''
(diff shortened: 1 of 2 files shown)


pylint_odoo/test_repo/broken_module/models/broken_model.py
```diff
@@ -46,6 +49,18 @@ class TestModel(models.Model):
         copy=True,
     )
 
+    # This is a inherit overwrite field then don't should show errors related
+    # with creation of fields.
+    field_state_overwrite = fields.Selection(
```
moylop260: This section is offtopic for this patch
Could you delete it, please?

JesusZapata: @moylop260 
Thanks!

pylint_odoo/checkers/no_modules.py
```diff
@@ -375,7 +375,9 @@ def colon_list_to_dict(self, colon_list):
                           'renamed-field-parameter'
                           )
     def visit_call(self, node):
-        if node.as_string().lower().startswith('fields.'):
+        if 'fields' == self.get_func_lib(node.func) and \
```
moylop260: Change `\` (ends) by `(` (starts)

JesusZapata: @moylop260 
Done

dreispt:
Does it have a lint check code number? I don't see it.
Can we add it to the docs (the README afaicr)?

moylop260:
The method visit_call is used for all check numbers related with declaration of fields.

This patch is better to detect fields declaration.
Before: all methods starting with "fields"
Now: all methods starting with fields and child of a class and assigned from root parameters of class.

PR #117, opened 2017-02-13
[FIX] xml-attribute-translatable: Skip if version is different 8.0

JesusZapata:
Active xml-attribute-translatable if the version is 8.0

Referring to https://github.com/Vauxoo/pylint-odoo/pull/111

### DIFF
--- a/pylint_odoo/checkers/modules_odoo.py
+++ b/pylint_odoo/checkers/modules_odoo.py
@@ -745,6 +745,9 @@ class ModuleChecker(misc.WrapperModuleChecker):
         """The xml attribute is missing the translation="off" tag
             Example  <attribute name="groups">sale.group</attribute>
         """
+        if (self.linter._all_options['valid_odoo_versions'].config
+                .valid_odoo_versions != ['8.0']):
+            return True
         self.msg_args = []
         for xml_file in self.filter_files_ext('xml', relpath=True):
             for record in self.get_xml_records(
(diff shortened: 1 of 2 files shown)


pylint_odoo/checkers/modules_odoo.py
```diff
@@ -745,6 +745,9 @@ def _check_xml_attribute_translatable(self):
         """The xml attribute is missing the translation="off" tag
             Example  <attribute name="groups">sale.group</attribute>
         """
+        if (self.linter._all_options['valid_odoo_versions'].config
+                .valid_odoo_versions != ['8.0']):
```
pedrobaeza: Isn't the contrary?

JesusZapata: @pedrobaeza 
Not, this check only available if it is version 8

pedrobaeza: Ok, then you need to change the title of the PR

JesusZapata: Okay, you were right

dreispt:
Can you fix the commit message?

JesusZapata:
@dreispt 
This commit https://github.com/OCA/pylint-odoo/pull/117/commits/2629d4c230609d15f9826a949f7ab42e847ef3a8 and the PR have the same description
What description should the commit have?

dreispt:
@JesusZapata It's OK - my bad.

Issue #132, opened 2017-06-15
How to disable a warning locally in XML file ?


adrienpeiffer:
Hello,

I have a case in which I would like to disable a constraint in a specific XML file. Is there a way to do that?

Thanks,

Adrien

pedrobaeza:
Can you specify which warning is bothering you?

adrienpeiffer:
Hi Pedro,

I have the following case :

```
<record id="dep1" model="hr.department"> <!-- I create the department-->
    <field name="name">dep1</field>
</record>

<record id="emp1" model="hr.employee"> <!-- I create the manager of the created department-->
    <field name="name">emp1</field>
    <field name="department_id" ref="dep1"/>
</record>

<record id="dep1" model="hr.department"> <!-- I assign the manager to the department-->
    <field name="manager_id" ref="emp1"/>
</record>
```

In this situation, I have this error : duplicate-xml-record-id)

Thanks,

Adrien

pedrobaeza:
The solution for avoiding the warning is to put the second definition in another file. The check is needed because it's a very usual error to copy/paste and don't change XML-ID.

adrienpeiffer:
@pedrobaeza This does not seem to work.

```
odoo/addons/module/demo/demo_manager.xml:19:0: W7902  Duplicate xml record id "demo/dep1_noupdate_0" in demo/demo.xml:40 (duplicate-xml-record-id)
```

pedrobaeza:
Ouch, I remembered bad how to bypass it. @moylop260 can you help?

moylop260:
I understand that this case could be a problem if we don't have a o2m for the m2o of the other model.

But for the case of `hr.department` there is a `member_ids` o2m field.

You could use:

```
<record id="emp1" model="hr.employee"> <!-- I create the manager of the created department-->
    <field name="name">emp1</field>
</record>

<record id="dep1" model="hr.department"> <!-- I create the department-->
    <field name="name">dep1</field>
    <field name="manager_id" ref="emp1"/>
    <field name="member_ids" eval="[(4, ref('emp1')]"/>
</record>

```


We will have this issue open to create a pylint-disable for xml files.
Thanks!

moylop260:
@JesusZapata 
Could you help me with this behaviour, please?

JesusZapata:
@moylop260 
Sure

PR #135, opened 2017-06-19
[IMP] pylint_odoo: Skip pylint check on one xml file

JesusZapata:
This PR adding the functionality to skip xml resource like `pylint: disable=CHECK` the `CHECK` should be the name of the check, is not supported the number of the pylint check

This functionality only support `xml` files, one example of use is 

```xml
<?xml version="1.0" encoding="utf-8"?>
<!-- pylint:disable=deprecated-data-xml-node -->
<odoo>
    <data>
        ...
    </data>
</odoo>
```

Fix https://github.com/OCA/pylint-odoo/issues/132

### DIFF
--- a/pylint_odoo/misc.py
+++ b/pylint_odoo/misc.py
@@ -3,6 +3,7 @@ import csv
 import os
 import re
 import subprocess
+import inspect
 
 from lxml import etree
 from pylint.checkers import BaseChecker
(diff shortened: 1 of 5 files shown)


JesusZapata:
@pedrobaeza @adrienpeiffer @moylop260 
Yours comments is welcome

sbidoul:
Very nice, thanks!

Issue #121, opened 2017-03-12
check 'website' manifest key is a valid URI


sbidoul:
pypi checks the url key is a valid uri according to rfc3986, therefore addons with an invalid website key in the manifest cannot be published to pypi.

Hence, I'm suggesting the addition of the following check to pylint-odoo, similar to [what pypi does](https://github.com/pypa/pypi-legacy/blob/5937ede003477561bc48a041384eeb22091d30d7/webui.py#L2699-L2709):

```python
import rfc3986

website = manifest.get('website')
uri = rfc3986.uri_reference(website)
if not uri.is_valid(require_scheme=True, require_authority=True):
    # error
if uri.scheme not in {"http", "https"}:
    # error
```

dreispt:
:+1:

moylop260:
@jesuszapata could you help us to add this new check, please?

JesusZapata:
@sbidoul 
Sorry for my long time answering

I try used `rfc3986` but this library not made the correct validation

![image](https://user-images.githubusercontent.com/1387970/27336820-b3ef0c5a-559e-11e7-81cb-6baf7312b780.png)

I proposal used the `validators` http://validators.readthedocs.io/ library 

![image](https://user-images.githubusercontent.com/1387970/27336906-f2c8018e-559e-11e7-92c3-0911a4eba7c1.png)

Are you agree ?

pedrobaeza:
Maybe first library "failed" because they are 2 valid URLs. Can you detect a comma and pass the validator?

sbidoul:
I think I saw the rf3986 used in pypi source code. We need to use the same
check as pypi. I'll check later.

moylop260:
@sbidoul 
Is possible use a built-in python library like as `urllib` to detect a valid url?
In order to reduce the packages to install

sbidoul:
The code sample in the issue description is the exact same used on pypi. So for me it's good enough to use that one. Note it does raise an error if there is a space after the comma. 

But I of course a stricter validation is always nice.

PR #136, opened 2017-06-20
[ADD] website-manifest-key-not-valid-uri: For validate if the website into manifest is valid URI

JesusZapata:
This modification validate the website key into the manifest file using the `rfc3986` library

If the website contains `,` is considered a url bad formatted

Fix https://github.com/OCA/pylint-odoo/issues/121

### DIFF
--- a/pylint_odoo/checkers/no_modules.py
+++ b/pylint_odoo/checkers/no_modules.py
@@ -56,6 +56,7 @@ import re
 import types
 
 import astroid
+import rfc3986
 from six import string_types
 from pylint.checkers import BaseChecker, utils
 from pylint.interfaces import IAstroidChecker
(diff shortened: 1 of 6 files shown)


moylop260:
@JesusZapata 
I saw many OCA modules with this case where more than one company contributor is contributing.
I'm not sure if this is a change that the board approval.

screen shot 2017-06-20 at 10 06 10

@gurneyalex @pedrobaeza @max3903 
How we should manage the website field from manifest file?

We should use oca website for manifest and our custom page from README similar to author?

pedrobaeza:
A correct URL is now mandatory for publishing to PyPi, so no problem in branches turning red. We must fix them. About the website to put, there's a debate in https://github.com/OCA/maintainer-tools/pull/276

pylint_odoo/checkers/no_modules.py
```diff
@@ -538,6 +545,17 @@ def visit_dict(self, node):
                 self.add_message('resource-not-exist', node=node,
                                  args=(key, resource))
 
+        # Check if the website is valid URI
+        website = manifest_dict.get('website', '')
```
moylop260: Reduce the cyclomatic complexity here a little using by default `manifest_dict.get('website') or 'https://'` and avoid nested `if`

JesusZapata: @moylop260 
I made reduction of cyclomatic complexity

JesusZapata:
@pedrobaeza 
In this PR skips the check if the website have `,` 
Example `https://odoo-community.org,https://odoo.com` is skipped

This is what you mean by this comment https://github.com/OCA/pylint-odoo/issues/121#issuecomment-309765331 ?

If the condition is right, Could you approve this PR ?

PR #138, opened 2017-07-06
[IMP] don't report xml files referenced in report tag as unreferenced

hbrunn:
this supports xml files being pulled by some `<report ... xml="$file" xsl="$file" />` tag


moylop260:
Good point!
Thanks!

Do you have planned add a test case for this matter?

hbrunn:
no, but i should. setting to work in progress until i find time for that

moylop260:
@JesusZapata 
Could you help us to add tests for this check?

JesusZapata:
@hbrunn 
I go to made the test of this PR, could you give me an example when on the label `report` has `<report ... xml="$file" xsl="$file" />`
While more specific better

hbrunn:
thank you so much for picking up my uncleaned garbage. Everything satisfying https://github.com/OCA/OCB/blob/10.0/odoo/import_xml.rng#L70 should be taken into account, but https://github.com/OCA/OCB/blob/8.0/addons/account/account_report.xml#L4 contains some specific examples

JesusZapata:
@moylop260 
This PR now has test for this funcionality 
And travis is on green

PR #140, opened 2017-07-28
[IMP] missing-readme: Check for other supported file names

IOAyman:
For instance, README with .rst, .md or .txt extensions
These filenames are supported by Odoo when loading module info

_**ref:**_
`load_information_from_description_file():`
   https://github.com/odoo/odoo/blob/e6863851/odoo/modules/module.py#L337
   https://github.com/odoo/odoo/blob/e6863851/odoo/modules/module.py#L28

### DIFF
--- a/pylint_odoo/checkers/modules_odoo.py
+++ b/pylint_odoo/checkers/modules_odoo.py
@@ -139,6 +139,7 @@ ODOO_MSGS = {
 
 DFTL_README_TMPL_URL = 'https://github.com/OCA/maintainer-tools' + \
     '/blob/master/template/module/README.rst'
+DFTL_README_FILES = ['README.rst', 'README.md', 'README.txt']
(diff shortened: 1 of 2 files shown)


pylint_odoo/checkers/modules_odoo.py
```diff
@@ -404,11 +405,14 @@ def _check_rst_syntax_error(self):
         return True
 
     def _check_missing_readme(self):
-        """Check if exists ./README.rst file
+        """Check if exists ./README.{rst,md,txt} file
         :return: If exists return True else False
         """
         self.msg_args = (self.config.readme_template_url,)
-        return os.path.isfile(os.path.join(self.module_path, 'README.rst'))
+        for f in DFTL_README_FILE:
```
moylop260: Use a variable with a better name, please f is shortest

PR #142, opened 2017-08-02
[REF] cache: Cache other extension files just if is a odoo module

moylop260:
Clean [TODO comment](https://github.com/OCA/pylint-odoo/blob/128d018791d2e74b6fbd9c5db19b0024c7f9e92b/pylint_odoo/misc.py#L99)

Currently if we have a big python project using pylint-odoo (non odoo module) is too slow because try get all files of the project.


moylop260:
@JesusZapata What do you think about?

moylop260:
Fast-tracking since that is:
 - A small fix
 - tests are green
 - It was a planned TODO
 - Big performance change

Issue #145, opened 2017-08-23
[RFC] A computed method shouldn't use write


moylop260:
Use `self.update` instead of `self.write` for computed methods in order to avoid permission errors

moylop260:
I won't working on this anymore

PR #148, opened 2017-08-29
[ADD] computed-method-should-not-use-write: A computed method shouldn't use write

JesusZapata:
**WIP**

This new check named `computed-method-should-not-use-write` detects the call to the method `write` inside the computed method

This detects two variants:
- Call directly the method `write`.
```
...
self.write()
...
```
- Transient call to `write` method
```
...
for item in self:
    item.write()
...
```

Fix https://github.com/OCA/pylint-odoo/issues/145

### DIFF
--- a/pylint_odoo/checkers/no_modules.py
+++ b/pylint_odoo/checkers/no_modules.py
@@ -364,7 +364,9 @@ class NoModuleChecker(BaseChecker):
         }),
     )
 
-    _name_computed_methods = []
+    def __init__(self, *args):
+        self._name_computed_methods = []
+        BaseChecker.__init__(self, *args)
 
     def open(self):
         super(NoModuleChecker, self).open()
--- a/pylint_odoo/checkers/no_modules.py
+++ b/pylint_odoo/checkers/no_modules.py
@@ -364,6 +364,8 @@ class NoModuleChecker(BaseChecker):
         }),
     )
 
+    _name_computed_methods = []
+
     def open(self):
         super(NoModuleChecker, self).open()
         self.config.deprecated_field_parameters = \
@@ -429,6 +431,9 @@ class NoModuleChecker(BaseChecker):
                 if isinstance(argument, astroid.Keyword):
                     argument_aux = argument.value
                     deprecated = self.config.deprecated_field_parameters
+                    if argument.arg == 'compute' and argument.value.value:
+                        self._name_computed_methods.append(
+                            argument.value.value)
                     if argument.arg in ['compute', 'search', 'inverse'] and \
                             isinstance(argument_aux, astroid.Const) and \
                             isinstance(argument_aux.value, string_types) and \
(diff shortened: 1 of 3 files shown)


JesusZapata:
@moylop260 
You comment is welcome

pylint_odoo/checkers/no_modules.py
```diff
@@ -568,6 +576,28 @@ def visit_functiondef(self, node):
         if not node.is_method():
             return
 
+        if node.name.startswith('_compute_'):
```
moylop260: What about use `fields.*(compute='THIS_PARAMETER')`?
In order to search all valid computed fields with/without standard

JesusZapata: @moylop260 
I refactor my previous code to detect the `fields.*(compute='THIS_PARAMETER')`
I I added various code of example to test this new check

JesusZapata:
@moylop260 
Do you could check this PR ?

pylint_odoo/checkers/no_modules.py
```diff
@@ -358,6 +364,8 @@ class NoModuleChecker(BaseChecker):
         }),
     )
 
+    _name_computed_methods = []
```
moylop260: If you use this variable when is affected and is not cleaned... so the variable is changed.

I mean,
screen shot 2017-09-08 at 18 28 51


Better you should use a object variable like as `self.variable`

JesusZapata: @moylop260 
Thanks for teach me this behaviour, I fixed my little mistake

pylint_odoo/test_repo/broken_module/models/broken_model.py
```diff
@@ -173,6 +173,25 @@ def method_date(self):
         help="Many 2 Many"
     )
 
+    def _compute_name(self):
```
moylop260: How this method is identified as computed if there is not defined from `meth(compute='compute_method')`?

JesusZapata: The computed method `_compute_name`  is defined in

https://github.com/OCA/pylint-odoo/blob/0a6242ba4ad720a867e8e23953732aa568c97329/pylint_odoo/test_repo/broken_module/models/broken_model.py#L28-L35

And the method `my_method_compute` is defined in

https://github.com/OCA/pylint-odoo/blob/0a6242ba4ad720a867e8e23953732aa568c97329/pylint_odoo/test_repo/broken_module/models/broken_model.py#L38-L44

I just use the existing code to avoid created another code to test this funcionality

JesusZapata:
@moylop260 
This PR now is green and I answered all question related with the changed

moylop260:
Using this PR and running 
```bash
git clone https://github.com/vauxoo/addons-vauxoo.git -b 8.0 --single-branch --depth=1 /tmp/addons-vauxoo
cd /tmp/addons-vauxoo
pylint --load-plugins=pylint_odoo -d all -rn -e odoolint *
```

I get the following error:
```bash
File "/private/tmp/py27_pylint_pr148/lib/python2.7/site-packages/pylint_odoo/checkers/no_modules.py", line 485, in visit_call
    if (item.iter.name == 'self' and
AttributeError: 'Subscript' object has no attribute 'name'
```

For the following line of code:
https://github.com/Vauxoo/addons-vauxoo/blob/3b847e930b03dc558be48e7d971c644f8e3fe98e/account_aged_partner_balance_vw/wizard/webkit_parser_header_fix.py#L147

(Maybe you like reproduce this error from tests before to fix it)

moylop260:
Closing since that there are not new comments since a while

Issue #146, opened 2017-08-24
[RFC] Deprecated <tree> attributes


naglis:
`string` attribute in `<tree>` node is [deprecated since 8.0](https://github.com/odoo/odoo/blob/420e433fa3a6b671868fa578e6006dd215570706/doc/reference/views.rst#lists) (no longer displayed) and `colors` and `fonts` are [deprecated since 9.0](https://github.com/odoo/odoo/blob/420e433fa3a6b671868fa578e6006dd215570706/doc/reference/views.rst#lists) in favor of `decoration-{$name}`.

I went ahead an implemented a crude version of it, however, there is the question if something like this would likely be integrated into _pylint-odoo_?

If yes, then I have a few further questions:

1. Since the rule deals with certain versions of Odoo only, the check should be skipped in other versions in order to avoid false positives. I saw [`valid_odoo_versions` config value being used](https://github.com/OCA/pylint-odoo/blob/d0b0d100057666f21bbb7b84c1779e2c02ca8c80/pylint_odoo/checkers/modules_odoo.py#L770-L771) in one place. Is this a good/right approach?
2. What is the policy for comming up with new message codes? Do I simply increment the highest existing code by one?

moylop260:
> valid_odoo_version

That approach looks good to me.
We will avoid use one pylint configuration file for each odoo version and will avoid many false positives/negatives if you don't have a specified configuration file.

> Do I simply increment the highest existing code by one?

There is not a policy. Just not duplicated and not WIP code used from PR.

PR #149, opened 2017-08-30
[ADD] xml-deprecated-tree-attribute: Implements check for deprecated tree attributes

naglis:
Implements check for deprecated XML `<tree>` node attributes (`colors`, `fonts` and `string`) as stated in #146.

### DIFF
--- a/pylint_odoo/checkers/modules_odoo.py
+++ b/pylint_odoo/checkers/modules_odoo.py
@@ -128,6 +128,11 @@ ODOO_MSGS = {
         'xml-attribute-translatable',
         settings.DESC_DFLT
     ),
+    'W%d42' % settings.BASE_OMODULE_ID: (
+        '%s Deprecated <tree> xml attribute "%s"',
+        'xml-deprecated-tree-attribute',
+        settings.DESC_DFLT
+    ),
     'W%d39' % settings.BASE_OMODULE_ID: (
         '%s Use <odoo> instead of <odoo><data> or use <odoo noupdate="1">'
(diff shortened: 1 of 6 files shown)


pylint_odoo/checkers/modules_odoo.py
```diff
@@ -810,3 +825,49 @@ def _check_xml_attribute_translatable(self):
...
+                'xpath': './/tree[@string]',
+            },
+        ]
+        valid_versions = set(
+            self.linter._all_options['valid_odoo_versions'].config
+            .valid_odoo_versions)
+
+        def check_is_applicable(check):
+            return (check['attr'] in self.config.deprecated_tree_attributes and
+                    bool(valid_versions - check['skip_versions']))
```
moylop260: What about use the `__manifest__` version value for the module?

naglis: Do you know if there is some helper to get the version from `__manifest__`?

Also, what if the version is invalid, eg. does not contain Odoo version (`1.0.0` or even `8.0.0` where `8` is the major version of the module)?

naglis:
Hi @moylop260,

> Could you help us with the new guideline about https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md please?

Sure. Wondering where to put the new guideline - add some new section under [OCA Guidelines -> XML files](https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#xml-files)?

moylop260:
Hi @naglis 
Yes, that is the correct side

PR #150, opened 2017-08-31
[FIX] misc: TypeError argument of type 'NoneType' is not iterable

moylop260:


moylop260:
@naglis Could you review this PR, plase?

naglis:
Hi @moylop260,

Nice catch!

How about adding a regression test case for this?

I see `record.get` is repeated three times. Might be worth reducing the repetitiveness, eg.:

```python
ref = record.get('id', '')
xml_module, xml_id = ref.split('.') if '.' in ref else ['', ref]

```

Other than that, LGTM 👍 


On a slightly related note, while reading the docstring of `get_xml_redundant_module_name`:
https://github.com/OCA/pylint-odoo/blob/7f1bcffd514fa63498d6820b41d3481c4fa1af5f/pylint_odoo/misc.py#L387-L402
I have noticed that:
- The docstring has a typo - it describes the `model` param, where in the function definition it is `module`. I guess it is a blind C/P mistake from [`get_xml_records`](https://github.com/OCA/pylint-odoo/blob/master/pylint_odoo/misc.py#L354-L356)
- The `model` param description does not match method behavior - it says that _if model is None then get all_. However, from the code it is evident that `xml_ids` is only appended to when `module` is set, so we get an empty list (and not all existing ids) when `module` is not set.

JesusZapata:
@naglis 
I made your change proposed 
Thanks for your detail on the time to review :+1:

moylop260:
@pedrobaeza could you merge it?

PR #151, opened 2017-09-04
[REF] jslintrc: Disable prefer-spread

moylop260:
Discussion about: https://github.com/OCA/web/pull/708#issuecomment-326538915
cc @yajo @chienandalu


moylop260:
@pedrobaeza Could you merge it, please?

Issue #152, opened 2017-09-04
[RFC] Tighten valid module version check


naglis:
Currently, the following module versions will be considered as valid by _pylint-odoo_ when using default configuration:

- `10+0.1.0.0`, because the dot from `valid_odoo_versions` is not escaped when building the valid module version pattern (and the `.` inside a regex matches any character other than a newline): https://github.com/OCA/pylint-odoo/blob/a14ddca6f0d57e603c55f60b0128b129ad5fb81e/pylint_odoo/checkers/no_modules.py#L657-L661
- `10.0.1.0.0WHATEVER`, because `re.match` [is used](https://github.com/OCA/pylint-odoo/blob/a14ddca6f0d57e603c55f60b0128b129ad5fb81e/pylint_odoo/checkers/no_modules.py#L661) inside `formatversion()`, which [matches the string from the beginning](https://docs.python.org/2/library/re.html#re.match), however, there is no `$` at the end of the default value for `manifest_version_format` to match the end of the string: https://github.com/OCA/pylint-odoo/blob/a14ddca6f0d57e603c55f60b0128b129ad5fb81e/pylint_odoo/checkers/no_modules.py#L237
Thus, anything can follow once the version is matched from the beginning of the string.

I have started working on a fix, however, I have stumbled into a slight problem for which I need feedback from the authors - if a _.pylintrc_ file is used with `manifest_version_format` defined inside it using the `%(valid_odoo_versions)s` placeholder, then `manifest_version_format` will be set automatically by `ConfigParser` (which [pylint uses for configuration parsing](https://github.com/PyCQA/pylint/blob/e231a409a83a23c24c8b7dbbaca79314eee6d4c6/pylint/config.py#L462)) due to [interpolation](https://docs.python.org/2/library/configparser.html#ConfigParser.ConfigParser). This means that at the point when the `manifest_version_format_parsed` string is built: https://github.com/OCA/pylint-odoo/blob/a14ddca6f0d57e603c55f60b0128b129ad5fb81e/pylint_odoo/checkers/no_modules.py#L657-L661
[...]

PR #153, opened 2017-09-04
[FIX] manifest-version-format: Use real dot from regex

moylop260:
Fix https://github.com/OCA/pylint-odoo/issues/152

cc @naglis

### DIFF
--- a/pylint_odoo/checkers/no_modules.py
+++ b/pylint_odoo/checkers/no_modules.py
@@ -234,7 +234,7 @@ DFTL_METHOD_REQUIRED_SUPER = [
 DFTL_VALID_ODOO_VERSIONS = [
     '4.2', '5.0', '6.0', '6.1', '7.0', '8.0', '9.0', '10.0'
 ]
-DFTL_MANIFEST_VERSION_FORMAT = r"(%(valid_odoo_versions)s)\.\d+\.\d+\.\d+$"
+DFTL_MANIFEST_VERSION_FORMAT = r"({valid_odoo_versions})\.\d+\.\d+\.\d+$"
 DFTL_CURSOR_EXPR = [
     'self.env.cr', 'self._cr',  # new api
     'self.cr',  # controllers and test
@@ -328,7 +328,7 @@ class NoModuleChecker(BaseChecker):
             'metavar': '<string>',
             'default': DFTL_MANIFEST_VERSION_FORMAT,
             'help': 'Regex to check version format in manifest file. '
-            'Use "%(valid_odoo_versions)s" to check the parameter of '
+            'Use "{valid_odoo_versions}" to check the parameter of '
             '"valid_odoo_versions"'
         }),
         ('cursor_expr', {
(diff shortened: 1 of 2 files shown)


naglis:
LGTM from technical point of view, however, I find that now it becomes a bit ambiguous on how to define the valid versions in the config file - I think I will find myself writing `8.0` instead of `8\.0`.

Also, I think it breaks the existing version checks inside _pylint_odoo_ checkers, eg.:

https://github.com/OCA/pylint-odoo/blob/42418ab69664e06c0b834016d667e88ab970f99e/pylint_odoo/checkers/modules_odoo.py#L800-L801

And [the ones](https://github.com/versada/pylint-odoo/blob/a8b57fc4c1b04a89ad77c61c2d95ff347446f203/pylint_odoo/checkers/modules_odoo.py#L854-L857) from #146.

moylop260:
@naglis you have a good point.
Good catch!
Thank you

Could you review the new changes, please?

naglis:
@moylop260 nice, did not know about `re.escape` :+1:

However, now the issue with string interpolation in `ConfigParser` which I have described in #152 remains. For example, if I have the following `.pylintrc` file and use it with _pylint-odoo_:
```ini
[ODOOLINT]
manifest_version_format=(%(valid_odoo_versions)s)\.\d+\.\d+\.\d+$
valid_odoo_versions=10.0
```
Then due to the string interpolation in `ConfigParser` by the time [`formatversion()`](https://github.com/OCA/pylint-odoo/blob/a14ddca6f0d57e603c55f60b0128b129ad5fb81e/pylint_odoo/checkers/no_modules.py#L657) is called `self.config.manifest_version_format` will already be formatted to `(10.0)\.\d+\.\d+\.\d+$` and the `re.escape`'d version numbers will simply be ignored by the `%` function (non-existent placeholders do not raise an exception). So version numbers like `10+0.1.0.0` will still be able to slip by with such a config file :/

moylop260:
I got it.
Thank you

I can reproduce it using:

`pylint --rcfile=pylint_manifest_version.cfg --load-plugins=pylint_odoo pylint_odoo/test_repo/broken_module`

```conf
[MESSAGES CONTROL]
disable=all
enable=manifest-version-format

[ODOOLINT]
manifest_version_format=(%(valid_odoo_versions)s)\.\d+\.\d+\.\d+
valid_odoo_versions=10.0
```

 - screen shot 2017-09-05 at 10 20 36

The `manifest_version_format` is using magical interpolation previously like as your comment: https://github.com/OCA/pylint-odoo/issues/152

Let me see if there is a method to pre-process the parameter from definition

moylop260:
You won man
We need bypass the magic interpolation then using other parsing string like as the "hated" `format` is fine for this case 😃 

Thank you @naglis for your patient to understand the issue

naglis:
@moylop260 no problem, it's a tricky one to explain. Thank you for your work on this issue 🙇

moylop260:
Could you give me your approval, please?

naglis:
I am not a reviewer, but you have my approval 👍

moylop260:
Could you review @lasley @pedrobaeza @dreispt @JesusZapata please?

pylint_odoo/checkers/no_modules.py
```diff
@@ -655,9 +655,10 @@ def camelize(self, string):
         return re.sub(r"(?:^|_)(.)", lambda m: m.group(1).upper(), string)
 
     def formatversion(self, string):
-        self.config.manifest_version_format_parsed = \
-            self.config.manifest_version_format % dict(
-                valid_odoo_versions='|'.join(self.config.valid_odoo_versions))
+        self.config.manifest_version_format_parsed = (
+            self.config.manifest_version_format.format(
+                valid_odoo_versions='|'.join(
+                    map(re.escape, self.config.valid_odoo_versions))))
```
lasley: Can we use some intermediary variables here? I'm having a bit of trouble following all the nested method calls IMO

moylop260:
@lasley 
What about now?

PR #154, opened 2017-09-04
[FIX] disable xml check: get method called when the method is not called directly

JesusZapata:
This change is neccesary because the funcionality to skip the xml check detect the method than called the another method `_skip_files_ext`

The way I found to know the method than called was the package [inspect](https://docs.python.org/2/library/inspect.html)

https://github.com/OCA/pylint-odoo/blob/42418ab69664e06c0b834016d667e88ab970f99e/pylint_odoo/misc.py#L233-L236

Related with this another PR https://github.com/Vauxoo/pylint-odoo/pull/126


JesusZapata:
@moylop260 
In this PR contains the fix when the method `filter_files_ext` is not called directly, for example in https://github.com/OCA/pylint-odoo/blob/42418ab69664e06c0b834016d667e88ab970f99e/pylint_odoo/checkers/modules_odoo.py#L783

naglis:
LGTM other than the nitpick that the new test XML file should be called `skip_file_not_used.xml`, I guess.

This is a cool feature, I've been looking for the possibility to disable rule checks inside XML (when it did not exist) and ended up having to disable the conflicting rule in the CI. Many thanks @JesusZapata! 

Maybe we should have this documented somewhere, so it would be a bit easier to find, perhaps in the README?

JesusZapata:
@naglis 
Thanks for you comment
I fix your comment in my last commit

PR #155, opened 2017-09-05
[FIX] manifest-version-format: Adds matching for string end to default manifest version format

naglis:
This adds `$` to the end of default manifest version format to disallow versions like `10.0.1.0.0WHATEVER`.

Related: #152

### DIFF
--- a/pylint_odoo/checkers/no_modules.py
+++ b/pylint_odoo/checkers/no_modules.py
@@ -234,7 +234,7 @@ DFTL_METHOD_REQUIRED_SUPER = [
 DFTL_VALID_ODOO_VERSIONS = [
     '4.2', '5.0', '6.0', '6.1', '7.0', '8.0', '9.0', '10.0'
 ]
-DFTL_MANIFEST_VERSION_FORMAT = r"(%(valid_odoo_versions)s)\.\d+\.\d+\.\d+"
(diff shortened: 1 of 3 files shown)


pylint_odoo/checkers/no_modules.py
```diff
@@ -234,7 +234,7 @@
 DFTL_VALID_ODOO_VERSIONS = [
     '4.2', '5.0', '6.0', '6.1', '7.0', '8.0', '9.0', '10.0'
 ]
-DFTL_MANIFEST_VERSION_FORMAT = r"(%(valid_odoo_versions)s)\.\d+\.\d+\.\d+"
+DFTL_MANIFEST_VERSION_FORMAT = r"(%(valid_odoo_versions)s)\.\d+\.\d+\.\d+$"
```
moylop260: Add `^` to "starts with" too

naglis: `re.match()`, which [is used](https://github.com/OCA/pylint-odoo/blob/a14ddca6f0d57e603c55f60b0128b129ad5fb81e/pylint_odoo/checkers/no_modules.py#L661) in `formatversion()` to check the version number [checks from beginning of the string](https://docs.python.org/2/library/re.html#re.match), so the `^` is redundant.

moylop260: Oh yeah!
You are right! (again!) 😄

PR #157, opened 2017-09-16
[ADD] xml-deprecated-qweb-directive: Implements check for deprecated QWeb directives

naglis:
This adds a check for use of deprecated QWeb directives `t-*-options` (`t-field-options`, `t-esc-options` and `t-raw-options`) which are [to be removed after v10](https://github.com/odoo/odoo/blob/8f99b24f6cb1ea70b371e2944ff36b75a6f9c80e/odoo/addons/base/ir/ir_qweb/ir_qweb.py#L155) and should be replaced with the `t-options` directive.

Currently, you only get a warning in the logs as the templates which use these directives are compiled, but [only in dev mode](https://github.com/odoo/odoo/blob/8f99b24f6cb1ea70b371e2944ff36b75a6f9c80e/odoo/addons/base/ir/ir_qweb/ir_qweb.py#L160-L161) (`--dev` CLI parameter).

I haven't found any usage of `t-raw-options` in Odoo or elsewhere on GitHub, however, it [is theoretically possible](https://github.com/odoo/odoo/blob/8f99b24f6cb1ea70b371e2944ff36b75a6f9c80e/odoo/addons/base/ir/ir_qweb/qweb.py#L1030), so I have included it as well.

### DIFF
--- a/pylint_odoo/checkers/modules_odoo.py
+++ b/pylint_odoo/checkers/modules_odoo.py
@@ -134,6 +134,11 @@ ODOO_MSGS = {
         'xml-deprecated-tree-attribute',
         settings.DESC_DFLT
     ),
+    'W%d43' % settings.BASE_OMODULE_ID: (
+        '%s Deprecated QWeb directive "%s". Use "t-options" instead',
+        'xml-deprecated-qweb-directive',
+        settings.DESC_DFLT
+    ),
     'W%d39' % settings.BASE_OMODULE_ID: (
         '%s Use <odoo> instead of <odoo><data> or use <odoo noupdate="1">'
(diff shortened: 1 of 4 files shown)


moylop260:
Thanks
Could you create PR for guideline from https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md, please?

naglis:
@moylop260, sure, I'll make a PR with a new guideline.

moylop260:
Hello @naglis 
Could you send me a email, please?

I like to proposal you for [PSC](https://odoo-community.org/page/ogranization) of this project

moylop260:
Merging since that the guideline was approved

PR #159, opened 2017-10-02
[REF] checkers: Open file with string in order to avoid decodes

moylop260:


pylint_odoo/test_repo/broken_module/xml_special_char.xml
```diff
@@ -2,7 +2,7 @@
 <openerp>
     <data noupdate="1">
         <record id="res_country_mx" model="res.country">
-            <field name="name">México</field>
+            <field name="name">M�xico</field>
```
moylop260: NOTE: We are using a non-utf8 char in order to reproduce the unicode errores

JesusZapata:
@moylop260 
Now your pull request is green

I ran this version of `pylint_odoo` over the `odoo/addons` and now ran well

moylop260:
@lepistone could you help us with a review, please?

PR #161, opened 2017-10-14
Added ArchLinux package reference

hmrodrigues:
Adds reference to ArchLinux package in AUR (Arch User Repository)

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -126,6 +126,10 @@ Or
 
 ``# pip install --upgrade --pre pylint-odoo``
 
+Or
+
+If using ArchLinux, AUR package python-pylint-odoo
+
 Usage
 -----
 


moylop260:
Thanks @hmrodrigues 

Registering this package in pypi we used a [OCA](https://odoo-community.org) official user/token.

What about use a OCA user to register it in https://aur.archlinux.org/packages/python-pylint-odoo ?

cc @max3903 @gurneyalex @pedrobaeza

pedrobaeza:
OK, can be a good option. What do we need?

hmrodrigues:
@pedrobaeza You need to create a account and then create a Orphan Request so the AUR guys transfer the package to the OCA account.

If you want, I can maintain the package after the transfer, you just need to add me as a co-maintainer.

moylop260:
Closing since that there are not new comments since a while

Issue #160, opened 2017-10-03
Warning about Version arguments in file resource path "?v=1"


a-da:
![selection_010](https://user-images.githubusercontent.com/17334646/31135649-eed7d29c-a865-11e7-8f50-37d4359d6f43.png)

This will not fail in *debug=* mode, but will fail in normal mode.

We spend 1 hour do understand the reason.

We can add a check as  

    "?" in file_path or "#" in file_path

moylop260:
Could you help me a litlle more to understand it?

a-da:
This is the point, its hard to understand, what warning is trying to say.

It try to open css fille open("xxx_affiliation_widget/static/widget.css?v=1") and fail.

```
<?xml version="1.0" encoding="UTF-8"?>
<openerp>
    <data>
        <template id="assets_backend" name="xxx_affiliation_widget" inherit_id="web.assets_backend">
            <xpath expr="." position="inside">
                <!-- wrong but is working in odoo web debug mode -->
                <link rel="stylesheet"
                      href="/xxx_affiliation_widget/static/widget.css?v=1"/>

               <!-- correct but is working working odoo web debug mode also -->
                <link rel="stylesheet"
                      href="/xxx_affiliation_widget/static/widget.css"/>

                <script type="text/javascript"
                        src="/xxx_affiliation_widget/static/widget.js"/>
            </xpath>
        </template>
    </data>
</openerp>
```

![selection_015](https://user-images.githubusercontent.com/17334646/31270338-4f113340-aa84-11e7-8651-ca045739a5bf.png)

moylop260:
I got it
Thank you for the clarification.

+1
-- 
Moisés López Calderón
Vauxoo - OpenERP's Gold Partner
Mobile: (+521) 477-752-22-30
Office: (+52) 477-773-33-46
web: http://www.vauxoo.com
twitter: @vauxoo
           @moylop260
hangout: moylop260@vauxoo.com

moylop260:
@JesusZapata 
Could you add a new check, if the `href` of a link is a valid file path?

moylop260:
@a-da Could you test if https://github.com/OCA/pylint-odoo/pull/163 makes the work?

a-da:
@moylop260
tested, working

```
virtualenv -p python2 /tmp/a
/tmp/a/bin/pip install git+https://github.com/vauxoo-dev/pylint-odoo.git@master-oca-character-not-valid-in-resource-link-jesuszapata
[...]

PR #163, opened 2017-10-19
[ADD] character-not-valid-in-resource-link: To valid the character '?/#' into the in src/href

JesusZapata:
This pull request check if the `<script src>` or `<link href>` contains the character `?` or `#`  and if this a valid absolute path

Fix https://github.com/OCA/pylint-odoo/issues/160#issuecomment-336995763


JesusZapata:
@moylop260
Your opinion is welcome

JesusZapata:
@moylop260 
I change the way to calculate the bad character on the resource path and I add more test cases

JesusZapata:
@moylop260 
I made small change to guarantee the Windows OS compatibility

PR #165, opened 2017-10-26
[ADD] Add licence-allowed: OEEL-1

acsonefho:
- Add licence OEEL-1 into allowed-licence.
Can be useful is somebody create a custom module to extend an Enterprise module.


max3903:
I am sorry guys. No OEEL modules in the OCA. Only OSI approved licenses please.

Please revert.

If a module depends on an enterprise module, LGPL is fine.

pedrobaeza:
@max3903 this tools can be used by other projects outside OCA, so shouldn't it accept it?

max3903:
Ok, so can we add all the other licenses and in the configuration file of the OCA installation, set that only OSI-approved ones should be allowed.

pedrobaeza:
@moylop260 can this be done?

moylop260:
FYI https://github.com/OCA/maintainer-quality-tools/pull/502

Issue #164, opened 2017-10-23
Disable "no-utf8-coding-comment" check for odoo >= v11.0


moylop260:
Obeying @Yajo https://github.com/OCA/maintainer-tools/pull/304 😄 

1. Disable check `no-utf8-coding-comment` for >= v11.0
2. New check `unnecessary-utf8-coding-comment` for >=v11.0

PR #168, opened 2017-10-30
[ADD] unnecessary-utf8-coding-comment: The utf8 comment is unnecessary if the version of odoo is 11.0

JesusZapata:
And in this commit the version of odoo is readed from the manifest file

For example the new module `eleven_module` is one module to test this functionality this is a module with the version 11.0 in the manifest file, and when I run the `pylint_odoo` over this module `pylint_odoo` only detect the check referent a odoo 11.0 this is because into the class of check has a new attribute `msgs_odoo_version`

```python
...
msgs_odoo_version = {
    'no-utf8-coding-comment': ['4.2', '5.0', '6.0', '6.1', '7.0', '8.0', '9.0', '10.0'],
    'unnecessary-utf8-coding-comment': ['11.0'],
}
...
```
In the `msgs_odoo_version` should be put the name of the check and the version supported by this check

Fix https://github.com/OCA/pylint-odoo/issues/164


JesusZapata:
@moylop260 
Your opinion is welcome.

moylop260:
What about use `maxversion` and `minversion` style like as original pylint uses for python version but this time for odoo module version?

 - [pylint/checkers/stdlib.py#L90 - maxversion](https://github.com/PyCQA/pylint/blob/dd78aedcf0f499188ad8232f2193ba62c791d33e/pylint/checkers/stdlib.py#L90)
 - [pylint/checkers/base.py#L447 - minversion](https://github.com/PyCQA/pylint/blob/dd78aedcf0f499188ad8232f2193ba62c791d33e/pylint/checkers/base.py#L447)

JesusZapata:
@moylop260 
After to test the new change I can say that way not work the conf value `minversion` and `maxversion` is only to check the python [version](
https://github.com/PyCQA/pylint/blob/56a6723f96dddd1e70e668d808e8446fd53b841f/pylint/utils.py#L193) 

```python
        if self.minversion is not None and self.minversion > sys.version_info:
            return False
        if self.maxversion is not None and self.maxversion <= sys.version_info:
            return False
```
And this change cause the two check with the parameter `maxversion` and `minversion` not emit check because the python version is not recognized

image

I try to create another conf value called `odoo_minversion` and `odoo_maxversion` this way failure because this configuration is not recognized for `pylint`

The source code

image

The mistake
image

The `pylint` only recognized the `minversion` and `maxversion` [specifically](https://github.com/PyCQA/pylint/blob/56a6723f96dddd1e70e668d808e8446fd53b841f/pylint/utils.py#L174-L175
)
```python
    def __init__(self, checker, msgid, msg, descr, symbol, scope,
                 minversion=None, maxversion=None, old_names=None):
```

moylop260:
Ok, we can't use it from definition of the check because the method don't receive extra parameters.

So we will need a additional dict to add odoo_min_version and odoo_max_version, something like as


```python
odoo_check_versions = {
    'check-name': {
        'max_odoo_version': (10, 0),
        'min_odoo_version': (8, 0),
    },
}
```

And process it

JesusZapata:
@moylop260 
I change the way to detect the version supported by the check now the way to put the version supporte is the following

```python
odoo_check_versions = {
    'check-name': {
        'max_odoo_version': (10, 0),
        'min_odoo_version': (8, 0),
    },
}
```

pylint_odoo/test_repo/broken_module/__openerp__.py
```diff
@@ -4,7 +4,7 @@
     # missing license
     'author': 'Many People',  # Missing oca author
     'description': 'Should be a README.rst file',
-    'version': '8_0.1.0.0',
+    'version': '8.0.1.0.0',
```
moylop260: We need this wrong case tested

pylint_odoo/misc.py
```diff
@@ -44,34 +48,26 @@ def join_node_args_kwargs(node):
...
-        """Get manifest file path
-        :param node_file: String with full path of a python module file.
-        :return: Full path of manifest file if exists else return None"""
-        if os.path.basename(node_file) == '__init__.py':
-            for manifest_basename in settings.MANIFEST_FILES:
-                manifest_file = os.path.join(
-                    os.path.dirname(node_file), manifest_basename)
-                if os.path.isfile(manifest_file):
-                    return manifest_file
+    def set_caches(self):
```
moylop260: Could you reduce the changes to minimal ones?
I mean, the method `set_caches` was changed of position from line 99 to line 61 but is not need it

pylint_odoo/checkers/no_modules.py
```diff
@@ -231,9 +231,6 @@
     'create', 'write', 'read', 'unlink', 'copy',
     'setUp', 'setUpClass', 'tearDown', 'default_get',
 ]
-DFTL_VALID_ODOO_VERSIONS = [
```
moylop260: Why you changed this global variable to `misc`?

pylint_odoo/misc.py
```diff
@@ -172,6 +165,62 @@ def _check_{NAME_KEY}(self, module_path)
...
+                return
+        return super(PylintOdooChecker, self).add_message(*args, **kwargs)
+
+    def _is_version_supported(self, version, name_check):
+
+        def version_to_tuple(version):
+            if '.' not in version and version.count('.') != 1:
+                return version
+            version = version.split('.')
+            return tuple([int(version[0]), int(version[1])])
```
moylop260: What about use `packaging.version.parse("8.0") == packaging.version.parse("8.0.1.0.1"[:3]) ## Result: True`?
And change the definition of valid versions `min_odoo_version` and `max_odoo_version ` to string instead of tuples.

pylint_odoo/checkers/format.py
```diff
@@ -41,13 +44,20 @@
...
     name = settings.CFG_SECTION
     msgs = ODOO_MSGS
+    odoo_check_versions = {
+        'no-utf8-coding-comment': {
+            'min_odoo_version': (4, 2),
+            'max_odoo_version': (10, 0),
+        },
+        'unnecessary-utf8-coding-comment': {
+            'min_odoo_version': (11, 0),
+            'max_odoo_version': (11, 0),
```
moylop260: For `12.0` apply this same check so you should define just `min_odoo_version` for this case.

pylint_odoo/checkers/format.py
```diff
@@ -17,6 +15,11 @@
         'no-utf8-coding-comment',
         settings.DESC_DFLT
     ),
+    'C%d02' % settings.BASE_FORMAT_ID: (
+        'UTF-8 coding is not necessary',
+        'unnecessary-utf8-coding-comment',
+        settings.DESC_DFLT
```
moylop260: Use original python version tuple format here too for the cases where a non-odoo-module is checked with this plugin

pylint_odoo/checkers/format.py
```diff
@@ -17,6 +15,11 @@
         'no-utf8-coding-comment',
         settings.DESC_DFLT
```
moylop260: Use original python version tuple format here too...

pylint_odoo/test_repo/eleven_module/__manifest__.py
```diff
@@ -0,0 +1,12 @@
+{
+    'name': 'Eleven module for tests',
+    'license': 'AGPL-3',
+    'author': u'Jesus, Odoo Community Association (OCA)',
+    'version': '11.0.1.0.0',
```
moylop260: Create the module twelve too, please

pylint_odoo/checkers/no_modules.py
JesusZapata: @moylop260 
I try to used the `DFTL_VALID_ODOO_VERSIONS` since `no_modules.py` but being inside the file `misc.py` I try to import the `DFTL_VALID_ODOO_VERSIONS` but this import fail, below the output

```python
(Pdb) from .checkers.no_modules import DFTL_VALID_ODOO_VERSIONS
*** ImportError: cannot import name 'DFTL_MANIFEST_DATA_KEYS'
```
This trouble is for the import `from .modules_odoo import DFTL_MANIFEST_DATA_KEYS`

For that reason I moved this definition

pylint_odoo/test_repo/eleven_module/__manifest__.py
JesusZapata: Done, I created the module twelve_module

pylint_odoo/misc.py
JesusZapata: Done, I changed the definition of the version from tuple to string, and include the package `distutils.version.LooseVersion` to made more easy the comparation

pylint_odoo/test_repo/broken_module/__openerp__.py
JesusZapata: Done, changed revert

pylint_odoo/checkers/format.py
JesusZapata: Done, I deleted the max_odoo_version and keep only the min_odoo_version

pylint_odoo/misc.py
JesusZapata: Done, I keep only the change necessary

pylint_odoo/misc.py
```diff
@@ -172,6 +166,55 @@ def _check_{NAME_KEY}(self, module_path)
                         node.file = node_file_original
                         node.lineno = node_lineno_original
 
+    def visit_module(self, node):
+        self.wrapper_visit_module(node)
+
+    def add_message(self, *args, **kwargs):
+        version = (self.manifest_dict.get('version')
+                   if isinstance(self.manifest_dict, dict) else '')
+        match = re.match(r"(?P<version>\d+.\d+)\.\d+\.\d+\.\d+$", version)
```
moylop260: Here you can re-use global variables: `DFTL_MANIFEST_VERSION_FORMAT.format(valid_odoo_versions=version)`

moylop260: In fact, if the module don't have the good version standard the second criteria is use the `valid_odoo_versions` from config parameter.

I mean,
```python
valid_version = self.conf.valid_odoo_versions  # from pylint parameter
if match and len(name_check) >= 1:
    valid_version = match.group('version')  # from manifest
```

pylint_odoo/misc.py
```diff
@@ -172,6 +166,55 @@ def _check_{NAME_KEY}(self, module_path)
...
+            return True
+        odoo_check_versions = self.odoo_check_versions.get(name_check, {})
+        if not odoo_check_versions:
+            return True
+        version = LooseVersion(version)
+        min_odoo_version = LooseVersion(odoo_check_versions.get(
+            'min_odoo_version', DFTL_VALID_ODOO_VERSIONS[0]))
+        max_odoo_version = LooseVersion(odoo_check_versions.get(
+            'max_odoo_version', DFTL_VALID_ODOO_VERSIONS[-1]))
+        return (version >= min_odoo_version and version <= max_odoo_version)
```
moylop260: What about use a `between` style python: `min_odoo_version <= version <= max_odoo_version`?

pylint_odoo/misc.py
JesusZapata: I try to used the regular expresion defined in `DFTL_MANIFEST_VERSION_FORMAT` but this regular expresion not work as we need

```python
(Pdb) re.match(r"(?P<version>\d+.\d+)\.\d+\.\d+\.\d+$", '11.0.1.0.0').groups()
('11.0',)
(Pdb) re.match(r"({valid_odoo_versions})\.\d+\.\d+\.\d+$", '11.0.1.0.0')
```

pylint_odoo/misc.py
JesusZapata: Done, thanks for the trick about the python between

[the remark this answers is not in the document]
JesusZapata: @moylop260 

Yes, if the check not have a `min_odoo_version` and `max_odoo_version` then the check ever run

```python
def _is_version_supported(self, version, name_check):
    if not hasattr(self, 'odoo_check_versions'):
        return True 
```

For the right functioning the check should be `adding odoo_check_versions` and provide the `min_odoo_version` or `max_odoo_version`

```python
odoo_check_versions = {
    'no-utf8-coding-comment': {
        'min_odoo_version': '4.2',
        'max_odoo_version': '10.0',
    }
 }
```

pylint_odoo/misc.py
JesusZapata: I try to used the config to get the default default version but not working

```python
(Pdb) self.conf.valid_odoo_versions
*** AttributeError: 'FormatChecker' object has no attribute 'conf'
```

JesusZapata:
@moylop260 
All your comment was fixed or answered

pylint_odoo/misc.py
```diff
@@ -172,6 +167,65 @@ def _check_{NAME_KEY}(self, module_path)
                         node.file = node_file_original
                         node.lineno = node_lineno_original
 
+    def visit_module(self, node):
+        self.wrapper_visit_module(node)
+
+    def add_message(self, *args, **kwargs):
+        valid_odoo_versions = self.linter._all_options[
+            'valid_odoo_versions'].config.valid_odoo_versions
+        version = (self.manifest_dict.get('version')
```
moylop260: What about if version is not defined in the manifest?

pylint_odoo/misc.py
```diff
@@ -172,6 +168,63 @@ def _check_{NAME_KEY}(self, module_path)
...
+                   if isinstance(self.manifest_dict, dict) else '0')
+        version = LooseVersion(version)
+        short_version = '.'.join(map(str, version.version[:2]))
+        match = re.match(
+            DFTL_MANIFEST_VERSION_FORMAT.format(
+                valid_odoo_versions=short_version), version.vstring)
+        if not match:
+            valid_odoo_versions = self.linter._all_options[
+                'valid_odoo_versions'].config.valid_odoo_versions
+            version = LooseVersion(valid_odoo_versions[0] if
```
moylop260: version is not used anymore in the code below

pylint_odoo/misc.py
JesusZapata: Done, I changed the order to get the version

pylint_odoo/checkers/format.py
```diff
@@ -83,6 +92,10 @@ def process_tokens(self, tokens):
         if not tokens_identified.get(MAGIC_COMMENT_CODING_UTF8) and \
            not os.path.basename(self.linter.current_file) == '__init__.py':
             self.add_message('no-utf8-coding-comment', line=1)
+        if (tokens_identified.get(MAGIC_COMMENT_CODING_UTF8) and
```
moylop260: This `if` is the same to previous one
What about re-use it?

pylint_odoo/checkers/format.py
```diff
@@ -83,6 +92,10 @@ def process_tokens(self, tokens):
...
+                not os.path.basename(
+                    self.linter.current_file) == '__init__.py'):
```
moylop260: This message should be emitted for `__init__.py` files too

pylint_odoo/misc.py
JesusZapata: Your be right, I deleted the unnecessary code

pylint_odoo/misc.py
```diff
@@ -172,6 +178,62 @@ def _check_{NAME_KEY}(self, module_path)
...
+        version = (self.manifest_dict.get('version')
+                   if isinstance(self.manifest_dict, dict) else '0')
+        version = LooseVersion(version)
+        match = self.formatversion(version.vstring)
+        if not match:
+            valid_odoo_versions = self.linter._all_options[
+                'valid_odoo_versions'].config.valid_odoo_versions
+            version = LooseVersion(valid_odoo_versions[0] if
+                                   len(valid_odoo_versions) == 1 else '0')
+            match = self.formatversion(version.vstring)
```
moylop260: If you have a valid_odoo_version of length 1 so you have a short_version here
You don't need a re.match here again

pedrobaeza:
You should squash a bit the commit history next time.

moylop260:
@pedrobaeza 
FYI the commit history was squashed in one commit in the branch base.
https://github.com/OCA/pylint-odoo/commit/fc41ef8067c474f0cc5e27ec214745a1841392ad

pedrobaeza:
Ah, thank you very much for that. I thought it was merged as the message was that one.

Issue #170, opened 2017-11-27
W7936 : False positive error


rousseldenis:
In my tests folder, I've several test files and a common one.

The common one is partly imported like 'from .common import ...' in tests files.

The pylint script returns the W7936 error ('missing external dependency) on one test file.

I have to import it like 'from . import common'

PR #171, opened 2017-11-27
[FIX] pylint_odoo: Check if the package start with dot to skip

JesusZapata:
Check if the package start with dot to skip because this package is absolute import

This change now detect when the import start with dot for example

```python
from .no_exists import package
```
If the import start with dot that import is not detect as external dependency 

Related with https://github.com/Vauxoo/pylint-odoo/pull/134

Fix https://github.com/OCA/pylint-odoo/issues/170


JesusZapata:
@moylop260 
Your opinion is welcome

Issue #169, opened 2017-11-26
duplicate-xml-fields in search views


thomaspaulb:
The linter gives `duplicate-xml-fields` error in the case of this search view:

    <search>
        <field name="employee_id" />
        <field name="department_id" string="Department (subordinates)" filter_domain="[]" context="{...some context...}" />
        <field name="department_id" context="{...some other context...}" />
 
Duplicate fields should however be allowed in search views. Sometimes you just want two different filters on the same field.

PR #172, opened 2017-11-27
[FIX] pylint_odoo: The check duplicate-xml-fields skip if the field is inside the search tag

JesusZapata:
The check `duplicate-xml-fields` skip the comprobation if the field is inside the `search` tab

For example 

```xml
<record model="ir.ui.view" id="view_ir_config_search">
    <field name="model">ir.config_parameter</field>
    <field name="arch" type="xml">
        <search string="System Properties">
            <field name="key_config" context="{'one': 'data'}"/>
            <field name="key_config" context="{'two': 'data'}"/>
        </search>
    </field>
</record>
```

Fix https://github.com/OCA/pylint-odoo/issues/169

### DIFF
--- a/pylint_odoo/checkers/modules_odoo.py
+++ b/pylint_odoo/checkers/modules_odoo.py
@@ -571,23 +571,13 @@ class ModuleChecker(misc.WrapperModuleChecker):
             name = field.attrib.get('name')
             context = field.attrib.get('context')
             filter_domain = field.attrib.get('filter_domain')
-            sourceline = field.sourceline
             if not name:
                 continue
-            for in_field in fields:
-                if ((not in_field.attrib.get('name') or
-                     sourceline == in_field.sourceline) or
-                    (not
-                     (name == in_field.attrib.get('name') and
-                      context == in_field.attrib.get('context') and
-                      filter_domain == in_field.attrib.get('filter_domain')))):
-                    continue
-                all_fields.setdefault(
-                    (field.attrib.get('name'),
-                     field.getparent()), []).append(field)
+            all_fields.setdefault(
+                (name, context, filter_domain, field.getparent()), []).append(field)
         # Remove all keys which not duplicated by excluding them from the
-        return dict(((field_xml_name, parent_node), nodes) for
-                    (field_xml_name, parent_node), nodes in
+        return dict(((name, context, filter_domain, parent_node), nodes) for
+                    (name, context, filter_domain, parent_node), nodes in
                     all_fields.items() if len(nodes) >= 2)
 
     def _check_duplicate_xml_fields(self):
--- a/pylint_odoo/checkers/modules_odoo.py
+++ b/pylint_odoo/checkers/modules_odoo.py
@@ -568,14 +568,25 @@ class ModuleChecker(misc.WrapperModuleChecker):
         """
         all_fields = {}
         for field in fields:
-            field_xml = field.attrib.get('name')
-            if (not field_xml or (field.getparent() and
-                                  field.getparent().tag == 'search')):
+            name = field.attrib.get('name')
+            context = field.attrib.get('context')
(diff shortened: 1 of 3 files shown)


JesusZapata:
@thomaspaulb 
This change fix your issue https://github.com/OCA/pylint-odoo/issues/169 related with the `duplicate-xml-fields` check
Do you could check if this change fixed your problem ?

JesusZapata:
@moylop260 
Your opinion is welcome

pylint_odoo/checkers/modules_odoo.py
```diff
@@ -569,7 +569,8 @@ def _get_duplicate_xml_fields(self, fields):
         all_fields = {}
         for field in fields:
             field_xml = field.attrib.get('name')
-            if not field_xml:
+            if (not field_xml or (field.getparent() and
+                                  field.getparent().tag == 'search')):
```
moylop260: What about if is a inherited search view?

Shouldn't we skip for xpath node too?

JesusZapata: @moylop260 
You can show me the code to support this case to adding this code to the repository ?

moylop260: By example the following view is a inherited search view but don't have the tag `search` (Just use `xpath`):

https://github.com/odoo/odoo/blob/1f39b6021e08d471bc9f59162973e3297a808811/addons/pos_sale/views/sales_team_views.xml#L23

The main issue here is that if we skip `xpath` node too then we will have many false green results.

What about detect a duplicated field if there is same context and filter_domain?

| No | field name | context            | filter domain            |
|---|------------|--------------------|--------------------------|
| 1 | field1     | {}                 | []                       |
| 2 | field1     | {}                 | []                       |
| 3 | field1     | {'key1': 'value1'} | []                       |
| 4 | field1     | {'key1': 'value1'} | []                       |
| 5 | field1     | {}                 | [('field', '=' 'value')] |
| 6 | field1     | {}                 | [('field', '=' 'value')] |
| 7 | field1     | {'key1': 'value1'} | [('field', '=' 'value')] |

Duplicated:
 - 1 with 2
 - 3 with 4
 - 5 with 6

7 is not duplicated because have different context and domain

JesusZapata: @moylop260 
I thinks is more easy detect the name, context and the filter_doman 
I will made the changes necessary to detect this case

JesusZapata:
@moylop260 
I made the change and now the `field name`, `context` and `filter domain` is compared to check if the field is repeated

pylint_odoo/checkers/modules_odoo.py
```diff
@@ -568,13 +568,24 @@ def _get_duplicate_xml_fields(self, fields):
...
-            if not field_xml:
+            name = field.attrib.get('name')
+            context = field.attrib.get('context')
+            filter_domain = field.attrib.get('filter_domain')
+            sourceline = field.sourceline
+            if not name:
                 continue
-            all_fields.setdefault(
-                (field_xml, field.getparent()), []).append(field)
+            for in_field in fields:
```
moylop260: The following logic
```python
for field in fields:
    ...
   for in_field in fields:
```

It's so slow and so bad for performance.
Because you are running loops `n^n`

What about use a complex key for all_fields dict?

I mean,
```diff
- all_fields.setdefault((name, field.getparent()), []).append(field)
+ all_fields.setdefault((name, context, filter_domain, field.getparent()), []).append(field) 
```

If all keys values are duplicated so the items will be more than 2. (Like as original logic)

JesusZapata: @moylop260 
You're right again, Thanks for help me in this topic

JesusZapata:
@moylop260 
Now the code only have one `for`

PR #177, opened 2017-12-11
[DOC] README.rst: Adding documentation to the way to skip one check based on the Odoo Version

JesusZapata:
Adding documentation to the way to skip one check based on the Odoo Version using the `odoo_check_versions` variable

If you need to valid one check using the Odoo version you should be used the
new variable ``odoo_check_versions``

For example, if you have one new check `your-new-check` and you need run only
if the Odoo version is between 9.0 and 11.0, inside the class you should be
declare the follow :

```python
...
class FormatChecker(PylintOdooTokenChecker):
...
    odoo_check_versions = {
        'your-new-check': {
            'min_odoo_version': '9.0',
            'max_odoo_version': '10.0',
        }
    }
...

```

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -149,6 +149,27 @@ If you have external files you can add them in ``examples`` folder to skip.
 
 For rst-syntax-error skip unknown directives
 
+Skip one check based on the Odoo version
+----------------------------------------
+
+If you need to restrict a check to specific version(s) of Odoo, use the new variable ``odoo_check_versions``
+
+For example, if you have one new check `your-new-check` and you need to run it only
+if the Odoo version is between 9.0 and 11.0, inside the class you should declare the following :
+
+.. code-block:: python
+
+    ...
+    class FormatChecker(PylintOdooTokenChecker):
+    ...
+        odoo_check_versions = {
+            'your-new-check': {
+                'min_odoo_version': '9.0',
+                'max_odoo_version': '10.0',
+            }
+        }
+    ...
+
 Skip one xml check
 ------------------
 


JesusZapata:
@moylop260 
This PR is only to adding documentation to the new way to skip one check base on the Odoo version

JesusZapata:
Only one question, Why the documentation is `rst` and not `md`?

The `md` show better the formatted code related with `python` and `xml`

README.rst
```diff
@@ -147,6 +147,29 @@ If you have external files you can add them in ``examples`` folder to skip.
 
 For rst-syntax-error skip unknown directives
 
+Skip one check based on the Odoo version
+--------------
+
+If you need to valid one check using the Odoo version you should be used the
```
hbrunn: If you need to restrict a check to specific version(s) of Odoo, use the new variable ...

README.rst
```diff
@@ -147,6 +147,29 @@ If you have external files you can add them in ``examples`` folder to skip.
...
+new variable ``odoo_check_versions``
+
+For example, if you have one new check `your-new-check` and you need run only
```
hbrunn: ...you need to run it only

README.rst
```diff
@@ -147,6 +147,29 @@ If you have external files you can add them in ``examples`` folder to skip.
...
+if the Odoo version is between 9.0 and 11.0, inside the class you should be
```
hbrunn: ...you should add a declaration as follows...

hbrunn: or ..declare the following...

moylop260: Thanks for review... fixed

hbrunn:
should we add the extra commit to fix https://travis-ci.org/OCA/pylint-odoo/jobs/387938088#L699 here or in another PR?

moylop260:
Thanks for review @hbrunn 
I just fixed the rst syntax error.

Issue #176, opened 2017-12-11
The method message_post need the subject and body be translatable


JesusZapata:
The method `self.message_post` need the parameter `subject` and `body` be translatable

For example, the follow method raise the check `translation-required` because the parameter `subject` and `body` not is instance of `_` function

```python
self.message_post(
    subject='Error when trying to generate the invoice',  # translation-required
    body='<h2>%s</h2><hr>%s</hr>' % (  # translation-required
        _('Error when trying to generate the invoice'),
        _('The invoice should have a datetime set.')))
```

The correct way to this code is the follow

```python
self.message_post(
    subject=_('Error when trying to generate the invoice'),
    body=_('<h2>%s</h2><hr>%s</hr>' % (
        _('Error when trying to generate the invoice'),
        _('The invoice should have a datetime set.'))))
```

PR #178, opened 2017-12-13
[REF] The method message_post need the subject and body be translatable

JesusZapata:
The check `translation-required` now detects the parameter `subject` and `body` into the method `message_post`, those parameter should be translatable

This change detect when the parameter `subject` and `body` not is translatable, one example like the follow
```python
    self.message_post(subject='Subject not translatable',
        body='Body not translatable')
```

This change detects many way to not translatable string

```python
'Subject not translatable'
'Body not translatable %s' % variable
'Subject not translatable %(value)s' % {'value': variable}
'Subject not translatable {}'.format(variable)

```

Fix https://github.com/OCA/pylint-odoo/issues/176

### DIFF
--- a/pylint_odoo/checkers/no_modules.py
+++ b/pylint_odoo/checkers/no_modules.py
@@ -163,7 +163,7 @@ ODOO_MSGS = {
         settings.DESC_DFLT
     ),
     'C%d07' % settings.BASE_NOMODULE_ID: (
-        'String parameter of raise "%s" requires translation. Use _(%s)',
+        'String parameter on "%s" requires translation. Use %s_(%s)',
         'translation-required',
         settings.DESC_DFLT
     ),
@@ -392,7 +392,8 @@ class NoModuleChecker(misc.PylintOdooChecker):
                           'method-compute', 'method-search', 'method-inverse',
                           'sql-injection',
                           'attribute-string-redundant',
-                          'renamed-field-parameter'
+                          'renamed-field-parameter',
+                          'translation-required',
                           )
     def visit_call(self, node):
         if ('fields' == self.get_func_lib(node.func) and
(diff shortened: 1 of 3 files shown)


moylop260:
This message is a output to the user but many times we are missing translate it

@pedrobaeza @lasley

JesusZapata:
@moylop260 

I run this modification over the `Vauxoo/addons-vauxoo#11.0` and below is the output

```
(python2.7) odoo@a0f640a220c8:~/build/OCA/pylint-odoo$ python -m pylint -r n -d all -e translation-required --load-plugins=pylint_odoo /home/odoo/addons-vauxoo
No config file found, using default configuration
************* Module addons-vauxoo.send_author_mail.tests.test_mail_author
C: 55,22: String parameter on "message_post" requires translation. Use body=_('Message') (translation-required)
C: 85,22: String parameter on "message_post" requires translation. Use body=_('Message') (translation-required)
C:115,22: String parameter on "message_post" requires translation. Use body=_('Message') (translation-required)
C:145,22: String parameter on "message_post" requires translation. Use body=_('Message') (translation-required)
************* Module addons-vauxoo.purchase_third_validation.models.purchase_third_validation_installer
C: 66,12: String parameter on "Warning" requires translation. Use _('The limit amount to the second approbation must be less that amount to third validation') (translation-required)
************* Module addons-vauxoo.product_extended_variants.wizard.wizard_price
C: 38,15: String parameter on "message_post" requires translation. Use subject=_('Automatically Computed Standard Price') (translation-required)
C:133,12: String parameter on "message_post" requires translation. Use subject=_(('Updated at %s at %s') % ((IDENTIFIER, WHEN))) (translation-required)
C:139, 8: String parameter on "message_post" requires translation. Use subject=_(('Successfully updated at %s ') % (WHEN)) (translation-required)
C:139, 8: String parameter on "message_post" requires translation. Use body=_(('File is on %s for errors and in %s for Ok Products.') % ((logfname, logfull))) (translation-required)
```

Issue #180, opened 2017-12-15
Modify all check if that need specific Odoo version


JesusZapata:
Modify all check that depend of the Odoo version to use the new way to skip the check if that check fails with the Odoo version

For example the check `your-new-check` only run over the Odoo `9.0` and `10.0` within the class that defines the check `odoo_check_versions` to skip the check if the version is not equal to `9.0` or `10.0`

```python
...
class FormatChecker(PylintOdooTokenChecker):
...
    odoo_check_versions = {
        'your-new-check': {
            'min_odoo_version': '9.0',
            'max_odoo_version': '10.0',
        }
    }
...

```

moylop260:
Closing since that PR related is closed too.
If a check requires a particular odoo version we should open a new issue for that case.

PR #181, opened 2017-12-15
[REF] pylint_odoo: Modificar all check if that need specific Odoo version

JesusZapata:
Modify all check that depends on the Odoo version to use the new way to skip the check if that check fails with the Odoo version

The following check needs the Odoo version

- [x] xml-deprecated-qweb-directive
- [x] deprecated-openerp-xml-node
- [x] deprecated-data-xml-node

Fix https://github.com/OCA/pylint-odoo/issues/180

### DIFF
--- a/pylint_odoo/test/main.py
+++ b/pylint_odoo/test/main.py
@@ -144,7 +144,7 @@ class MainTest(unittest.TestCase):
         excluded_msgs = {
             'xml-attribute-translatable',
         }
-        extra_params = []
+        extra_params = ['--valid_odoo_versions=8.0']
         pylint_res = self.run_pylint(self.paths_modules, extra_params)
         msgs_found = pylint_res.linter.stats['by_msg'].keys()
         plugin_msgs = set(misc.get_plugin_msgs(pylint_res)) - excluded_msgs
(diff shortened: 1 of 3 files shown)


JesusZapata:
@moylop260 
In this PR I modify the `pylint-odoo` checks, to add the new way to check if the check applies to the version of Odoo using the `odoo_check_versions`

pylint_odoo/test/main.py
```diff
@@ -142,9 +142,9 @@ def test_25_checks_without_coverage(self):
         # Some messages can be excluded as they are only applied on certain
         # Odoo versions (not necessarily 8.0).
         excluded_msgs = {
-            'xml-deprecated-qweb-directive',
+            'xml-attribute-translatable',
         }
-        extra_params = ['--valid_odoo_versions=8.0']
+        extra_params = []
```
moylop260: why you remove this extra params?

JesusZapata: Sorry, not this change is not necessary, I only adding the deprecated-data-xml-node and openerp-exception-warning because this check not run over odoo 8.0

pylint_odoo/test/main.py
```diff
@@ -142,7 +142,9 @@ def test_25_checks_without_coverage(self):
         # Some messages can be excluded as they are only applied on certain
         # Odoo versions (not necessarily 8.0).
         excluded_msgs = {
-            'xml-deprecated-qweb-directive',
```
moylop260: And xml-deprecated-qweb-directive',?

moylop260:
Closing since that there are not new comments from a while

PR #182, opened 2017-12-19
[REM] Remove import guard check

lasley:
* Remove import guard check due to merge of odoo/odoo#14850

I feel like I'm missing something though. Comments/suggestions?


moylop260:
@JesusZapata 
thanks to https://github.com/odoo/odoo/commit/8226aa1db828d2a559c7ffaa31a27ef3e5ba4d0b
the check “missing-import-error” is not need anymore.
Could you disable it just for the stable version if is released?

moylop260:
Closing since that any comment was added after a while.

PR #184, opened 2018-01-11
[REF] jslintrc: Disable dot-location

moylop260:
Close https://github.com/OCA/maintainer-quality-tools/issues/524

### DIFF
--- a/pylint_odoo/examples/.jslintrc
+++ b/pylint_odoo/examples/.jslintrc
@@ -160,7 +160,7 @@
     "constructor-super": "error",
     "curly": "error",
     "default-case": "off",
-    "dot-location": "error",
+    "dot-location": ["error", "property"],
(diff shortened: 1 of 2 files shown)


pylint_odoo/examples/.jslintrc
```diff
@@ -160,7 +160,7 @@
     "constructor-super": "error",
     "curly": "error",
     "default-case": "off",
-    "dot-location": "error",
```
lasley: Can we switch to the [`property` style](https://eslint.org/docs/rules/dot-location#property) instead of completely disabling? IMO the lint is valid, just the default option is wrong.

pedrobaeza:
Travis is red

moylop260:
Now is green and show the real error from eslint if exists

PR #196, opened 2018-05-12
[FIX] translation-required: Fix some false positives

### LINKED ISSUE #187: Lint to check post messages are translatable is throwing false positives
The lint `translation-required` is intended to detect not-translatable messages withing methods like `message_post`, like in this code:
```
self.message_post(body='some message'))
```
And suggest using the function `_()` to make them translatable.

However, codes like the following are triggering false positives:
```
self.message_post(
    body=_('Error during the process: %s') % message,
    subtype='account.mt_invoice_validated')
```

Showing the following error message:
```
l10n_mx_edi_payslip/models/hr_payslip.py:264: [C8107(translation-required), HrPayslip.l10n_mx_edi_log_error] String parameter on "message_post" requires translation. Use body=_((_('Error during the process: %s')) % (message))
```

Not sure if this has anything to do with `message_post` and `_(...)` being located on different lines.

luisg123v:
This change causes the lint translation-required to take into account
certain cases which were triggering false positives, like the following:

`self.message_post(body=_('error: %s') % error_code)`

i.e. tose messages involving variables in the message string.

In addition, this adds some cases to the tests, to ensure they work as
expected.

Closes #187

### DIFF
--- a/pylint_odoo/checkers/no_modules.py
+++ b/pylint_odoo/checkers/no_modules.py
@@ -51,6 +51,7 @@ for more info visit pylint doc
 """
 
 import ast
+import itertools
 import os
 import re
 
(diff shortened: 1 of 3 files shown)


moylop260:
Excellent
Thank you!

I'll merge since that is a fix for a beta message not enabled for OCA projects yet

luisg123v:
thanks a lot! @moylop260

PR #197, opened 2018-05-19
[REF] .travis.yml: Allow both W503 and W504 formats

luisg123v:
This change allows both formatting styles when a line break is around a
binary operator, to be before or after the operator. This is due to a
change in PEP-8 recomendations.

For more info, see the following MQT issue:
https://github.com/OCA/maintainer-quality-tools/issues/545

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,7 +46,7 @@ install:
     - pip install restructuredtext_lint pygments
 
 script:
-    - flake8 --ignore=F601 --exclude=__init__.py .
+    - flake8 --ignore=F601,W503,W504 --exclude=__init__.py .
     - restructuredtext-lint ${TRAVIS_BUILD_DIR}/README.rst
     - tox -e $TOXENV,profile-stats
 


pedrobaeza:
But with this you are only changing the condition for checking this repository AFAIK, isn't it?

luisg123v:
@pedrobaeza 
Yes. This was already addressed in MQT, to modify the default configuration for flake8 tests (reported on https://github.com/OCA/maintainer-quality-tools/issues/545 and solved on https://github.com/OCA/maintainer-quality-tools/pull/546). However, this project doesn't depend on MQT, so that would need to be modified also from here.

luisg123v:
I think it might be a little bit of useful, because this project is based on a lot of conditions, and they are more readable when the line break is before the operator (e.g. https://github.com/OCA/pylint-odoo/pull/196)

luisg123v:
@moylop260 
Could you approve this, please?

PR #206, opened 2018-08-20
[FIX] missing-manifest-dependency: Exclude tests directory

luisg123v:
Since tests files are executed only when running tests, external
dependencies defined on those files should not be required in the
manifest as external Python dependencies. Those dependencies will not be
used in a non-testing environment and, if tests are run and dependencies
are not satisfied, tests will fail anyway.

This commit causes all files into the tests directory to be ignored by
the lint.

This is similar to: https://github.com/OCA/pylint-odoo/pull/75


luisg123v:
CC @moylop260

pylint_odoo/test/main.py
```diff
@@ -41,8 +41,8 @@
     'method-inverse': 1,
     'method-required-super': 8,
     'method-search': 1,
-    'missing-import-error': 4,
-    'missing-manifest-dependency': 2,
+    'missing-import-error': 7,
```
pedrobaeza: If we exclude test directory, why the number of errors raise?

luisg123v: Because we're also adding 6 new cases: 3 in the tests directory and 3 more in a model file. Then, there should be only 3 more cases.

luisg123v:
@moylop260  & @pedrobaeza 
This is not red anymore.

PR #209, opened 2018-08-21
Jslint jsdoc relax

Yajo:
## Relax some rules for jsdoc comments

This relaxation comes from the fact that sometimes you're extending another object, and thus you don't need to reexplain everything if it is explained in the parent class.

A common coding pattern is:

```js
return this._super.apply(this, arguments);
```

In such cases, it should be enough to declare the return type, since further descriptions should be found in the upstream class.

Another common pattern is reusing upstream parameters, or using parameters with names that are specific enough so as to not require a description, so the requirement of param description is relaxed here too.

## Some more coding styling

- Disable inline block spacing.
- Require multiline dangling commas.
- Enforce PEP8-like line limits.
- Require spaces before comments.
- Requre `"use strict"`.

---
@Tecnativa

### DIFF
--- a/pylint_odoo/examples/.jslintrc
+++ b/pylint_odoo/examples/.jslintrc
@@ -14,7 +14,7 @@
   "rules": {
     "no-alert": "error",
     "no-array-constructor": "error",
-    "no-bitwise": "error",
+    "no-bitwise": "off",
     "no-caller": "error",
     "no-case-declarations": "error",
     "no-catch-shadow": "error",
@@ -51,7 +51,9 @@
     "no-fallthrough": "error",
     "no-floating-decimal": "error",
     "no-func-assign": "error",
-    "no-implicit-coercion": "error",
+    "no-implicit-coercion": ["error", {
+      "allow": ["~"]
+    }],
     "no-implicit-globals": "error",
     "no-implied-eval": "error",
     "no-inline-comments": "error",
@@ -144,12 +146,19 @@
     "arrow-spacing": "off",
     "accessor-pairs": "error",
     "block-scoped-var": "off",
-    "block-spacing": "off",
+    "block-spacing": ["error", "always"],
     "brace-style": "error",
     "callback-return": "error",
     "camelcase": "off",
-    "comma-dangle": "off",
-    "comma-spacing": "off",
+    "capitalized-comments": ["error", "always", {
+      "ignoreConsecutiveComments": true,
+      "ignoreInlineComments": true
+    }],
+    "comma-dangle": ["error", "always-multiline"],
+    "comma-spacing": ["error", {
+      "before": false,
+      "after": true
+    }],
     "comma-style": "error",
     "complexity": [
       "error",
@@ -174,18 +183,22 @@
     "id-blacklist": "error",
     "id-length": "off",
     "id-match": "error",
-    "indent": "off",
+    "indent": "error",
     "init-declarations": "error",
     "jsx-quotes": "error",
     "key-spacing": "off",
-    "keyword-spacing": "off",
+    "keyword-spacing": "error",
     "linebreak-style": [
       "error",
       "unix"
     ],
-    "lines-around-comment": "off",
+    "lines-around-comment": "error",
     "max-depth": "error",
-    "max-len": "off",
+    "max-len": ["error", {
+      "code": 80,
+      "ignorePattern": "odoo\\.define\\(",
+      "tabWidth": 4
+    }],
     "max-lines": "off",
     "max-nested-callbacks": "error",
     "max-params": "off",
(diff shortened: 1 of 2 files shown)


pylint_odoo/examples/.jslintrc
```diff
@@ -185,7 +188,10 @@
     ],
     "lines-around-comment": "off",
     "max-depth": "error",
-    "max-len": "off",
+    "max-len": ["error", {
+      "code": 79,
```
pedrobaeza: Do we want to strict also here? I don't know if JS has the same limits as PEP8

pylint_odoo/examples/.jslintrc
```diff
@@ -215,23 +223,23 @@
     "quote-props": "off",
     "quotes": "off",
     "radix": "error",
-    "require-jsdoc": "off",
+    "require-jsdoc": "warn",
```
pedrobaeza: Can we detect which are the incorrect not documented instead of changing this to warn?

pylint_odoo/examples/.jslintrc
Yajo: We can set the limits we want. I'm open to suggestions, but I just wanna have a standard.

pylint_odoo/examples/.jslintrc
Yajo: The jsdoc parser is another option. This one raises a warn when there's a method without docstring. I don't wanna require it, but it's useful to have a reminder on the subject when you're writing a module.

pylint_odoo/examples/.jslintrc
pedrobaeza: Yeah, I'm asking if you know a convention on this for the JS part.

pylint_odoo/examples/.jslintrc
pedrobaeza: Then better to disable instead of having a lot of noise, but this is still useful in the case of new methods. That's why I ask if it's possible to distinguish them.

Yajo: I don't understand you... :sweat_smile: What do you want exactly and why?

pylint_odoo/examples/.jslintrc
Yajo: There are plenty of them, we just have to choose. The most common are 80 and 120 AFAIK.

Yajo:
Please go see https://github.com/OCA/web/pull/1022 to see an example of code that would complain 100% with these rules.

cc @moylop260

pylint_odoo/examples/.jslintrc
simahawk: personally I'd go 120 :)

pedrobaeza:
Why Travis fails?

moylop260:
@Yajo 
Thanks!
Could you increase the expected values, please?
 - https://github.com/OCA/pylint-odoo/blob/master/pylint_odoo/test/main.py#L33

@hugho-ad @oscarolar 
FYI

pylint_odoo/examples/.jslintrc
```diff
@@ -144,12 +146,15 @@
     "arrow-spacing": "off",
     "accessor-pairs": "error",
     "block-scoped-var": "off",
-    "block-spacing": "off",
-    "brace-style": "error",
+    "block-spacing": ["error", "never"],
+    "brace-style": "warn",
```
hugho-ad: Personally, I think this must be kept as an error in order to encourage the programmers to follow brace-style

pylint_odoo/examples/.jslintrc
hugho-ad: I'm with 80

Yajo: So what do we do? :thinking:

pedrobaeza: OK for me also for 80

pedrobaeza:
@Yajo maybe you should adapt the commit message before merging this with squash?

pylint_odoo/examples/.jslintrc
simahawk: I'll try to survive :sob: :rofl:

Yajo:
I'll do it now, thanks everyone.

BTW, when are we gonna start printing a :x: in travis when this fails? :thinking: 

Oh, and something we need to keep in mind: since future odoo v12 we're gonna be able finally to start using ES6+ features, so we'll need a different linter config for v12+ than for v11-.

pedrobaeza:
> since future odoo v12 we're gonna be able finally to start using ES6+ features

Are you sure about this? I heard that it's not yet possible.

Yajo:
BTW, ready to merge.

moylop260:
Could you check the red travis, please?
https://travis-ci.org/OCA/pylint-odoo/jobs/420010488#L2770-L2771

PR #210, opened 2018-08-22
[IMP] manifest-required-author: Allow multiple possible required authors

luisg123v:
There are cases when the required autor for modules may be one among
several ones, e.g. the author is required, but it is not always the
same. For cases like those, it is useful to support multiple values for
this lint.

This commit improves the lint so it accepts multiple values separated by
commas.

Test cases are provided to cover one and multiple values passed to the
parameter to set required authors.


luisg123v:
CC @moylop260

README.rst
```diff
@@ -11,7 +11,7 @@ Enable custom checks for Odoo modules.
...
 +-------+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+--------------------------------------+
-| C8101 | Missing author required "%s" in manifest file                                                                                                                                                                                                                                                      | manifest-required-author             |
+| C8101 | One of the following authors must be present in manifest: %s                                                                                                                                                                                                                                       | manifest-required-author             |
```
mreficent: Shouldn't this be `manifest-required-authors` instead?

luisg123v: I'd say no, because the lint wasn't renamed, only the parameter.

The reason for that is, the lint tells one required author is missing, while the parameter accepts multiple possible required authors. For that, I left the lint name singular and renamed the parameter to plural.

Of course, I'm open to suggestions regarding this.

rousseldenis:
@luisg123v There is problems now on already existing modules.

See https://travis-ci.org/OCA/stock-logistics-warehouse/jobs/420051631

rousseldenis:
@moylop260 @pedrobaeza

rousseldenis:
Seems that just one required author condition does not work

mreficent:
Maybe its because of the use of `"` instead of `'`.

pedrobaeza:
Yeah, I'm having the same problem. Let me check a bit, although it's @luisg123v the more experienced in this.

pedrobaeza:
OK, fixed in https://github.com/OCA/pylint-odoo/commit/f092c2370d9d59adb19c47665693df619f802b96

pedrobaeza:
I'm afraid it's not the good one. I continue checking.

pedrobaeza:
This is the good one!

https://github.com/OCA/maintainer-quality-tools/commit/c80d1d01bc5fe859f079dcc33b12f3644f1653d7

luisg123v:
@pedrobaeza  Thanks! I didn't noticed the difference because, in our cfg, we' were not using quotes

Issue #212, opened 2018-08-22
W7930(file-not-used) for CSV files into the 'migrations' folder


ebirbe:
I'm working in a project that has some CSV files into the `migrations` directory, and getting the next pylint error for those files:

![seleccion_076](https://user-images.githubusercontent.com/442938/44475050-0fcc6900-a602-11e8-85d7-e5630fe280c6.png)

These files are used only for the migration scripts, so I think is an issue to get that pylint message for that kind of files under the `migrations` directory.

/cc @moylop260

ebirbe:
It is similar, but not the same as https://github.com/Vauxoo/pylint-odoo/issues/133

PR #213, opened 2018-08-22
[IMP] file-not-used: Ignore unused files into the migrations directory

luisg123v:
If there are data files into the "migrations" directory, e.g. XML/CSV,
they are probably intended to be used by the migration script, hence
they  won't be referenced from the manifest.

This commit causes the lint to ignore files located into the
"migrations" directory.

Closes https://github.com/OCA/pylint-odoo/issues/212

### DIFF
--- a/pylint_odoo/checkers/modules_odoo.py
+++ b/pylint_odoo/checkers/modules_odoo.py
@@ -829,19 +829,17 @@ class ModuleChecker(misc.WrapperModuleChecker):
 
     def _check_file_not_used(self):
         """Check if a file is not used from manifest"""
-        self.msg_args = []
         module_files = set(self._get_module_files())
         referenced_files = set(self._get_manifest_referenced_files()).union(
             set(self._get_xml_referenced_files())
         )
-        for no_referenced_file in (module_files - referenced_files):
-            if (not no_referenced_file.startswith('static/') and
-                not (no_referenced_file.startswith('test/') or
-                     no_referenced_file.startswith('tests/'))):
-                self.msg_args.append((no_referenced_file,))
-        if self.msg_args:
-            return False
-        return True
+        excluded_dirs = ['static', 'test', 'tests', 'migrations']
+        no_referenced_files = [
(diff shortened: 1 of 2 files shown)


pylint_odoo/checkers/modules_odoo.py
```diff
@@ -829,19 +829,17 @@ def _get_module_files(self):
...
-            if (not no_referenced_file.startswith('static/') and
-                not (no_referenced_file.startswith('test/') or
-                     no_referenced_file.startswith('tests/'))):
-                self.msg_args.append((no_referenced_file,))
-        if self.msg_args:
-            return False
-        return True
+        excluded_dirs = ['static', 'test', 'tests', 'migrations']
+        no_referenced_files = [
+            file for file in (module_files - referenced_files)
```
moylop260: file is a reserverd keyword

pylint_odoo/checkers/modules_odoo.py
```diff
@@ -829,19 +829,17 @@ def _get_module_files(self):
...
+            if file.split('/')[0] not in excluded_dirs
```
moylop260: Use os.path methods

pylint_odoo/checkers/modules_odoo.py
moylop260: use a normal for to use a break early
or you can use sets too here?

pylint_odoo/checkers/modules_odoo.py
luisg123v: Done, used as much as I could. Since there's no way to extract the top directory from a path using `os.path`, I used `os.path.sep`

pylint_odoo/checkers/modules_odoo.py
luisg123v: - Done, changed the "file" variable. Thanks, good catch!
- I think an early break wouldn't be convenient in this case. Even though the method returns a boolean, it also stores all failing files and, if we break the loop, it could lead to non-detected failing files.

PR #215, opened 2018-09-25
[REF] tox: Upgrade new pylint for py3

moylop260:
* Old classes were removed for new versions and 1.9 is using a dual compatibility.:
  - https://github.com/PyCQA/astroid/blob/eaef0dbedb32ea9540505db6450454b5e8dbd10b/astroid/node_classes.py#L4261-L4267

* Remove python-lxml apt package since that new lxml is using wheels and travis is using pip cache

* Update compatibility with py3.7

* Use a dummy pylint configuration file in order to avoid using the default one from the HOME (Useful for local testing)

* pylint2 is not compatible with python<3 anymore so removing that cases from tox and travis

* pylint2 renamed `check_message_id` to `get_message_definition` in:
   - https://github.com/PyCQA/pylint/commit/5ccbf9eaa54c0c302c9180bdfb745566c16e416d

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,10 +1,14 @@
-astroid==1.6.3
-pylint==1.8.4
-pylint-plugin-utils==0.2.4
+astroid==1.6.5 ;python_version < '3'
+pylint==1.9.3 ;python_version < '3'
+
+astroid==2.0.4 ;python_version >= '3'
+pylint==2.1.1 ;python_version >= '3'
+
+pylint-plugin-utils==0.4
 docutils==0.14
-lxml>=2.3.2
+lxml>=4.2.3
 Pygments==2.2
-restructuredtext_lint==1.1
+restructuredtext_lint==1.1.3
 isort==4.3.4
 whichcraft
 rfc3986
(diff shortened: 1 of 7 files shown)


requirements.txt
```diff
@@ -1,10 +1,10 @@
-astroid==1.6.3
-pylint==1.8.4
+pylint==1.9.3 ;python_version < '3'
+pylint==2.1.1 ;python_version >= '3'
 pylint-plugin-utils==0.2.4
 docutils==0.14
-lxml>=2.3.2
+lxml==4.2.3
```
pedrobaeza: Is now this library faster to be compiled?

moylop260: This one has dual compatibility py2 and py3.7
Wheels are used for old and new version I think so

moylop260:
This PR is green and ready to review.

shepilov-vladislav:
@pedrobaeza Can you force to merge this pull?

moylop260:
Merging since that the unique comment is about time to compile `lxml` in the build.

Comparing the last stable travis build vs this PR build the time is not affected:
 - version

shepilov-vladislav:
Great ty @moylop260 ! You save my time!

PR #217, opened 2018-10-02
[IMP] file-not-used: Ignore unused files into the import directory

agajic-modoolar:
If there are data files into the "import" directory, e.g. XML/CSV,
they are probably intended to be used by the import python script, hence
they won't be referenced from the manifest.

This covers the case when there is complicated structure, or validation
that needs to be imported from modules init hook, and file is delivered
along with the module.

This commit causes the lint to ignore files located into the
"import" directory.

This is an extension of logics implemented in [commit](https://github.com/OCA/pylint-odoo/commit/a827027a939e20ed5591353216d163f08c88a930) made by @luisg123v.

cc @moylop260, @pedrobaeza: Please take a minute to review this small change. Thanks

### DIFF
--- a/pylint_odoo/checkers/modules_odoo.py
+++ b/pylint_odoo/checkers/modules_odoo.py
@@ -833,7 +833,7 @@ class ModuleChecker(misc.WrapperModuleChecker):
         referenced_files = set(self._get_manifest_referenced_files()).union(
             set(self._get_xml_referenced_files())
         )
-        excluded_dirs = ['static', 'test', 'tests', 'migrations']
+        excluded_dirs = ['static', 'test', 'tests', 'migrations', 'import']
         no_referenced_files = [
             f for f in (module_files - referenced_files)
             if f.split(os.path.sep)[0] not in excluded_dirs
--- a/pylint_odoo/test_repo/eleven_module/import/not_used_from_manifest.xml
+++ b/pylint_odoo/test_repo/eleven_module/import/not_used_from_manifest.xml
@@ -1,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+    <!-- Even though this file isn't referenced from the manifest, this shouldn't fail,
+        because it's into the "import" folder
+    -->
+</odoo>


luisg123v:
I like the idea of having a directory only for data imported/referenced from the code and not from the manifest. Has that directory been used before?

If not, I think it should be mentioned in the error message, so users may know about it when this lint is triggered. That way, they may decide if the involved file needs to be referenced from the manifest or moved to the "import" directory.

moylop260:
What about creates a configurated parameter for this kind of paths?
-- 
Moisés López Calderón
Mobile: (+521) 477-752-22-30
Twitter: @moylop260
hangout: moylop260@vauxoo.com
http://www.vauxoo.com - Odoo Gold Partner
Twitter: @vauxoo

agajic-modoolar:
What kind of config parameter you have in mind? Environment variable could do the job.
Adding configuration parameter just for this case would be inconsistent with other directories that could not be configured (i think of currently excluded directories that are hard coded).
We could agree on excluding common directories: 'static', 'test', 'tests', and use an environment variable for custom directories such as 'migrations', 'import'. This would impact modules that are currently passing without warning for having files in 'migrations' directory, but since it is still a fresh update, i think it would not make a big problem overall.
I will update PR according to that if we agree on proposed solution.

moylop260:
+1 for adding configuration parameter for extra non hard coded cases.
-- 
Moisés López Calderón
Mobile: (+521) 477-752-22-30
Twitter: @moylop260
hangout: moylop260@vauxoo.com
http://www.vauxoo.com - Odoo Gold Partner
Twitter: @vauxoo

agajic-modoolar:
@moylop260 You didn't mention "migrations" directory? I would also remove it from hardcoded list since it is not standard Odoo directory.

luisg123v:
@agajic-modoolar 
"migrations" is a standard Odoo directory, checked by Odoo when the version in the manifest as increased:

https://github.com/odoo/odoo/blob/23bec7fb7f8daf1252b85f22827112591d72818d/odoo/modules/migration.py#L54-L74

moylop260:
@agajic-modoolar

Feel free to re-open it when you have these parameters from configuration file:
 - https://github.com/OCA/pylint-odoo/pull/217#issuecomment-427007128

PR #220, opened 2018-10-31
Add OPL-1 to allowed licenses

katyukha:


katyukha:
Hello!

As i see [here](https://github.com/OCA/pylint-odoo/blob/8dcb1eb9df1e8623680f20f698d58b6e325a67de/pylint_odoo/checkers/no_modules.py#L227) `OPL-1` is not in allowed license list.

So that question is: Is there any reason that this license is not present in allowed list yet?

pedrobaeza:
@katyukha, OCA don't allow this kind of modules. If you use pylint for internal uses, just configure your CI to ignore this license check.

pedrobaeza:
Sorry, I'm seeing that the other private license is here. Reopening and checking.

pedrobaeza:
@moylop260 why private licenses are put here?

katyukha:
@pedrobaeza Thanks for fast answer)

moylop260:
@pedrobaeza
You are right.
We need remove private license from MQT conf file:

https://github.com/OCA/maintainer-quality-tools/blob/master/travis/cfg/travis_run_pylint_pr.cfg#L16

@katyukha
Here we are using just the most common licenses used from odoo sa, oca.

If you are using another one you can add it to your custom conf file like Pedro told us.

katyukha:
Thanks for answers. I am closing this PR.

Issue #222, opened 2018-11-13
Try-except lint check to be disabled in v12


pedrobaeza:
According commit https://github.com/odoo/odoo/commit/8226aa1db828d2a559c7ffaa31a27ef3e5ba4d0b, we don't need anymore to protect imports on 12.0, so we should remove the guideline for that version and don't make the check on 12.0 branches.

pedrobaeza:
cc @moylop260

pedrobaeza:
Maybe this should go in maintainer-quality-tools...

moylop260:
Thanks for reporting @pedrobaeza 
Could you check the following PR, please?
https://github.com/OCA/pylint-odoo/pull/223

PR #223, opened 2018-11-15
[REF] missing-import-error: Check enabled just for odoo <= 11.0

moylop260:
This check is deprecated since:
 - https://github.com/odoo/odoo/commit/8226aa1db828d2a559c7ffaa31a27ef3e5ba4d0b

Fix https://github.com/OCA/pylint-odoo/issues/222

cc @pedrobaeza @Yajo


pedrobaeza:
Check the change in guidelines in https://github.com/OCA/odoo-community.org/pull/33

Issue #225, opened 2018-11-16
sql-injection: consider case using `psycopg2.extensions.AsIs`


moylop260:
Steps:
1. Create a odoo module using the following sql injection case:
```python
from psycopg2.extensions import AsIs
self.env.cr.execute("SELECT id FROM %s", (AsIs('res_users; DELETE FROM res_users;'),))
```
2. Run pylint enabling sql-injection checker.

Saw:
sql-injection checker is not detected.

Expected:
sql-injection checker detected.

pedrobaeza:
Well, you are ruined the trick for avoiding the check, as there's no other way of for example build the SQL views reports.

moylop260:
We can use a sanitize method like
https://github.com/OCA/server-tools/issues/1426#issuecomment-439411355

Could you share me a code bypassing this check using AsIs?

In order to check if there is a way to skip a special case from the check

pedrobaeza:
This is one example: https://github.com/OCA/crm/blob/c0a0abc5c8b1a9ef17d224bbab768c8f089eddaf/crm_phonecall/report/crm_phonecall_report.py#L134

moylop260:
Using private attributes was fixed here:
 - https://github.com/OCA/pylint-odoo/pull/122/files

I mean, sql-injection is not face out if the variables startswith `_`

I just have created the following PR to test the case:
 - https://github.com/OCA/pylint-odoo/pull/226

pedrobaeza:
Can you try with a variable inside the same scope (not an object variable)?

moylop260:
Done
https://github.com/OCA/pylint-odoo/pull/226/files

pedrobaeza:
Well, I mean if it doesn't count when the variable is in the same scope. Any way, I don't like this, as you are reducing also the possibilities. What about if I want to put this:
```python
"CREATE VIEW %s AS (SELECT %s FROM res_partner)" % (_variable, self.select()))
```

moylop260:
I just have created a talbe of True with all cases using public and private attributes and using with and without underscored variables.

| Use                     | Result             |
|-------------------------|--------------------|
| self._private_attribute | skip sql-injection |
| self.public_attribute   | sql-injection      |
| _variable               | sql-injection      |
| variable                | sql-injection      |

Really, if you use a variable doesn't matter the name.
For OOP the attributes change if it starts with a underscore.
Then, the variables are out of scope for this case.

pedrobaeza:
[...]

PR #226, opened 2018-11-21
[REF] sql-injection: Test private attributes

moylop260:
Add a missing test case:
 - https://github.com/OCA/pylint-odoo/issues/225#issuecomment-440733022


moylop260:
@pedrobaeza 
Could we merge it, since that are valid test cases uncovered?

moylop260:
Merging since that the change is just to add unittests and it is approved

Issue #199, opened 2018-05-24
new check: development_status key


sbidoul:
Check that the `development_status` manifest key, if present, is one of the following strings

- Alpha
- Beta
- Production/Stable
- Mature

moylop260:
Check the following PR to enable the key as required:
 - https://github.com/OCA/maintainer-quality-tools/pull/582

PR #227, opened 2018-11-23
[ADD] development-status-allowed: New check to verify valid development status

moylop260:
Fix https://github.com/OCA/pylint-odoo/issues/199


moylop260:
@max3903 
Could you check it, please?

PR #241, opened 2019-03-08
[ADD] print-used: Use _logger instead of print

moylop260:
Since that new version of pylint is not considering a check for
"print()" in python3


moylop260:
I merged since that we have an approve
And we have the following issue with core `print-statement` check:
 - https://github.com/OCA/pylint-odoo/issues/243

Then we need this fix to disable core print-statement check and enable this one instead.

PR #244, opened 2019-05-17
Allow console statements

Yajo:
These are required in Odoo in order to mark a tour as failed manually.

It can also be helpful to log some interesting details for developers that shouldn't matter for users. Odoo uses it, in fact.

See https://github.com/OCA/website/pull/627#discussion_r285006941 where this was discovered.

### DIFF
--- a/pylint_odoo/examples/.jslintrc
+++ b/pylint_odoo/examples/.jslintrc
@@ -22,7 +22,7 @@
     "no-class-assign": "error",
     "no-cond-assign": "error",
     "no-confusing-arrow": "error",
-    "no-console": "error",
+    "no-console": "off",
     "no-const-assign": "error",
     "no-constant-condition": "error",
     "no-continue": "off",
--- a/pylint_odoo/test/main.py
+++ b/pylint_odoo/test/main.py
@@ -31,7 +31,7 @@ EXPECTED_ERRORS = {
     'file-not-used': 6,
     'incoherent-interpreter-exec-perm': 3,
     'invalid-commit': 4,
-    'javascript-lint': 25,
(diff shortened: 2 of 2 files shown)


pedrobaeza:
Isn't something like `raise` available in JS?

tbaden:
found this for js:
throw ... raises an exception in the current code block and causes it to exit, or to flow to next catch statement if raised in a try block.

console.error just prints out a red message to the browser developer tools javascript console and does not cause any changes of the execution flow

Yajo:
Yes, but that's not good for tours (if that's why you ask for it):

- [For Odoo 8-11, to fail a tour manually you have to do `console.log("error")`](https://github.com/odoo/odoo/blob/daf9763e6ecbaadb8f92b14fd2fadc310c54c690/odoo/tests/common.py#L310-L313).
- For Odoo 12, [additionally you can use `console.error()` for the same purpose](https://github.com/odoo/odoo/blob/2c3451de12bd120d0d72f7b50803f42e89f2f141/odoo/tests/common.py#L742-L744).

I'm not completely sure if it aborts the tour, or lets it finish but marks it as failed. In any case, we can't forbid using an Odoo feature just to satisfy a linter.

pedrobaeza:
Yes, the problem is the method used by Odoo here, but we have no choice. Please put that links as comments before the line for future references, and check Travis (is there any example file that is being linted in tests?)

Yajo:
Travis should go green now. I cannot add comments in JSON files, but the commit explains everything.

pedrobaeza:
Let's move on with this for avoiding the false linter.

PR #249, opened 2019-07-19
[REF] pylint_odoo: Supports odoo 13.0

moylop260:
Odoo version 13.0 is coming soon

TODO:
Fixing missing unnecessary-utf8-coding-comment:
 - `pylint_odoo/test_repo/broken_module3/__init__.py:1: [C8202(unnecessary-utf8-coding-comment), ] UTF-8 coding is not necessary`

NOTE: Testing locally using
`python3.7 -m pylint --load-plugins=pylint_odoo -d all -e unnecessary-utf8-coding-comment pylint_odoo/test_repo/broken_module3/__init__.py`

UPDATE:
The TODO will be worked from:
 - https://github.com/OCA/pylint-odoo/issues/252

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,6 @@ matrix:
 
 
 install:
-    - nvm install 11  # Update nodejs version to 11.latest, required by eslint
     # Remove packages installed from addons-apt-packages of travis file
     - sed -i '/node/d' ${TRAVIS_BUILD_DIR}/install.sh
     - sed -i '/pip install ./d' ${TRAVIS_BUILD_DIR}/install.sh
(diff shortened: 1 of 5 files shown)


moylop260:
IMHO If all our (OCA) forgot update new odoo version. It means that this project will need die.
In fact, I can add a regex to support valid possible version but the idea of this is just use a current valid number.

pedrobaeza:
OK, as you prefer. I always forget this part in the yearly new version cycle, but if you are there, then no problem :smiley:

moylop260:
FYI It is ready

PR #250, opened 2019-07-24
[ADD] method-default prefix

blaggacao:
For the same intantion behind search, compute, inverse naming convention
also alert on the same level for default. As being an indirectly called
method, too, this enables better static conde analysis.


moylop260:
LGTM 👍 
The travis red is offtopic for this PR since that stable branch is reproduced

Thanks!

blaggacao:
Ahhh, I had no internet on my way home (not at home), I wanted to block it. There is still one case I missed I guess... I'll check today and prepare a fix. Please don't release yet.

blaggacao:
But thanks for merging! :)

PR #256, opened 2019-09-12
[ADD] context-overridden: Better using kwargs instead of dictionary

moylop260:
More info about see docstring of the method:
 - https://github.com/odoo/odoo/blob/ff3064258a320e4008b3fb52a149e6751ae1e7f5/odoo/models.py#L5086-L5105

Using
  - `self.with_context({'key': 'value'})`

The original context is overridden

It should be:
  - `self.with_context(**{'key': 'value'})`
  - `self.with_context(key='value')`

If you need to override the full context only silent this check using:
  - `self.with_context({'key': 'value'})  # pylint: disable=context-overridden`
In order to allow an explicit overridden

Requested by @yajo from:
 - https://github.com/odoo/odoo/pull/36164

- [x] Add unittest

### DIFF
--- a/pylint_odoo/checkers/no_modules.py
+++ b/pylint_odoo/checkers/no_modules.py
@@ -245,6 +245,12 @@ ODOO_MSGS = {
         'translation-positional-used',
         settings.DESC_DFLT
     ),
+    'W%d21' % settings.BASE_NOMODULE_ID: (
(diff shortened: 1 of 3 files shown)


voronind:
:+1: 
Nobody cancel `disable`

PR #260, opened 2019-09-30
[IMP] jslintrc: Add support to ECMA ES6

Tardo:
Adds support to ECMA ES6

@Tecnativa

### DIFF
--- a/pylint_odoo/examples/.jslintrc
+++ b/pylint_odoo/examples/.jslintrc
@@ -285,5 +285,7 @@
     "yield-star-spacing": "off",
     "yoda": "error"
   },
-  "parserOptions": {}
+  "parserOptions": {
+    "ecmaVersion": 6,
+  }
 }


katyukha:
Hi,

It seems that Odoo 13 uses ES8 (async/await):
- https://github.com/odoo/odoo/blob/13.0/addons/web/static/src/js/views/abstract_controller.js#L206
- https://github.com/odoo/odoo/blob/13.0/addons/web_diagram/static/tests/diagram_tests.js

sbidoul:
cc/ @Yajo since you where discussing this in https://github.com/OCA/maintainer-quality-tools/pull/613

sbidoul:
I wonder how this relates to browser compatibility commitments of Odoo.

katyukha:
PS: For those who are looking for fast solution, it is possible to specify custom ESLint config with following params:

```bash
pylint --jslintrc=path/to/.jslintrc
```

katyukha:
It seems that ES8 is used mostly for internal UI (/web) and ES6 for website part.
(based on fast searching of `async`, `await` and `Promise` keywords in Odoo codebase)

pylint_odoo/examples/.jslintrc
```diff
@@ -285,5 +285,7 @@
     "yield-star-spacing": "off",
     "yoda": "error"
   },
-  "parserOptions": {}
+  "parserOptions": {
+    "ecmaVersion": 6,
```
Yajo: - Odoo 13 uses [`async/await` in normal code](https://github.com/odoo/odoo/blob/4ad7a99df0a25122cf6872f880e07b6148c82ea8/addons/web/static/src/js/views/abstract_controller.js#L206) (outside of tests), which is ES2016 (ES7)
- [Also uses the spread operator for browser code](https://github.com/odoo/odoo/blob/4ad7a99df0a25122cf6872f880e07b6148c82ea8/addons/website/static/src/js/content/snippets.animation.js#L693), which is ES2018 (ES9).

So, this should be:

```suggestion
    "ecmaVersion": 2018,
```

(It seems to me that year-versioning is easier to understand these days).

---

Note: [Year/version/features summary](https://en.wikipedia.org/wiki/ECMAScript).

katyukha:
May be it have sens to detect odoo version (for example via environment variables) and apply different configuration files for different odoo versions? For example enable es2018 only for odoo 13+ ?

Tardo:
Spread operator exists since ES2015 (ES6)
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax

Yajo:
True, I updated https://github.com/OCA/pylint-odoo/pull/260#discussion_r332952195

moylop260:
> @katyukha 
Maybe it has senses to detect odoo version (for example via environment variables) and apply different configuration files for different odoo versions? For example enable es2018 only for odoo 13+ ?


After black integration, each OCA repository will have a configuration file for each branch >=13.0
 - https://github.com/OCA/maintainer-quality-tools/pull/613

sbidoul:
Done in bb0ea41b12fba8b3b5e103c18bc7b6b4e1f16a39 for 13

PR #261, opened 2019-10-01
[FIX] rst-syntax-error: Using a 0 by default if line None is returned

moylop260:
cc @eilst 
FIX https://github.com/OCA/pylint-odoo/issues/262


moylop260:
/ocabot merge

Fast tracking since that it is a very small fix but it is raising error from other builds

PR #263, opened 2019-10-16
Fix incompatibility with isort configuration file

### LINKED ISSUE #247: Third party libraries such as psycopg2 marked as W7936(missing-manifest-dependency)
I tracked down an issue with a check returning erroneous result:

```
odoo/local-src/camptocamp_tools/utils.py:10: [W7936(missing-manifest-dependency), ] Missing external dependency "psycopg2" from manifest. More info: https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#external-dependencies
```

It happens here:

https://github.com/OCA/pylint-odoo/blob/67fa4893ce04b93e19612a577bc4563f0bfa62dc/pylint_odoo/checkers/modules_odoo.py#L356-L361

We add the list of modules to whitelist, psycopg2 is in the list, so we shouldn't have this warning. However, if you use `isort` to actually sort the imports, you may have a configuration file `.isort.cfg`. In this configuration, you will have the list of modules to consider as third party, so isort places them at the right place. Example:

```
[settings]
known_third_party = dateutil,openupgradelib,pkg_resources,psycopg2,requests,setuptools,urllib2,yaml
```

`isort.place_module()` builds a list of known patterns, and check first if the lib is a third party, returning this result first (if `psycopg2` is in the third party list and the std list, it will be considered as third party).

Refs (not in the most recent version, but easy to read): 
* https://github.com/timothycrosley/isort/blob/88a23e8aaacddf727ce01040dab31b9f7f55c497/isort/isort.py#L148-L155
* https://github.com/timothycrosley/isort/blob/88a23e8aaacddf727ce01040dab31b9f7f55c497/isort/isort.py#L257-L262

guewen:
The checker uses 'isort' to know if an imported library is in the
standard library or is a third party library. Then, for third party
libraries, the check will fail if it is not declared in the addon
manifest's external dependencies.

Some python modules are known to be used by Odoo and do not need to
be declared in the external dependencies. These libraries are listed
in a whitelist of the checker.

isort's SortImports class is initialized with an argument
'known_standard_library' where the whitelist is passed. It forces
isort to consider them as standard library, thus ignoring the check.

However, when the project's directory contains an '.isort.cfg' file,
because the project uses isort to actually sort the imports in the code,
the file usually contains a line such as:

[settings]
known_third_party = dateutil,openupgradelib,pkg_resources,psycopg2,requests,setuptools,urllib2,yaml

This line is necessary for isort to move the imports at the expected
place, after the standard imports.

When this file is present, the isort's 'place_module()' method will
consider it when called by the checker, and will consider the
whitelisted modules to be third party instead of the "expected" standard
library.

This commit directly checks if the module is in the whitelist, so we
avoid the confusion. isort is now only used to distinguish the standard
and third party which are not in the whitelist.

Fixes #247

### DIFF
--- a/pylint_odoo/checkers/modules_odoo.py
+++ b/pylint_odoo/checkers/modules_odoo.py
@@ -276,9 +276,18 @@ class ModuleChecker(misc.WrapperModuleChecker):
         node.file = self.linter.current_file
         self.inh_dup.setdefault(key, []).append(node)
 
+    def _build_whitelist_module_patterns(self):
+        known_patterns = []
+        for known_pattern in self.config.import_name_whitelist:
+            pattern = known_pattern.replace('*', '.*').replace('?', '.?')
+            known_patterns.append(re.compile('^' + pattern + '$'))
+        return known_patterns
+
     def open(self):
         """Define variables to use cache"""
         self.inh_dup = {}
+        patterns = self._build_whitelist_module_patterns()
+        self._whitelist_module_patterns = patterns
         super(ModuleChecker, self).open()
 
     def close(self):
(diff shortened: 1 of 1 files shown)


sbidoul:
@guewen BTW, do you know that Odoo 13 [now allows referencing PyPI distribution names in python depends](https://github.com/odoo/odoo/pull/25549) rather than the import name? It's the recommended way to do it. 

Sorry don't know to review this PR right now, I don't know if this changes something or not.

guewen:
Yes, and this checker doesn't take this into account. A similar check than https://github.com/odoo/odoo/pull/25549 should be added in 
https://github.com/OCA/pylint-odoo/blob/d8bf90dafc1038c602c52608007f3ef433241d7c/pylint_odoo/checkers/modules_odoo.py#L379-L380

Should be the matter of another PR

PR #268, opened 2019-12-27
WIP: [FIX] sql-injection: Supports multiple private args

moylop260:
Now ignores execute("..." % (self._table, ...))
and execute("...".format(table=self._table)) or execute("...".format(self._table))"

Since that it is private argument (no user input) and it is a valid
option to execute queries without sql-injection

Fixing the following valid case of use:
 - https://github.com/OCA/l10n-usa/pull/51/files#diff-1003d0a8ef76ca3b90e7581aae489425R57


moylop260:
It was already fixe from:
 - https://github.com/OCA/pylint-odoo/pull/344

Issue #281, opened 2020-04-25
Pylint 2.4.4 support


zhaqenl:
In light of [this issue](https://github.com/flycheck/flycheck/issues/1746#issuecomment-619356555) regarding the linter displaying output in HTML (which is not expected) in Pylint 2.3.0, is there any news regarding support for Pylint 2.4.4?

Installing `pylint-odoo` via `pip install pylint-odoo` reverts my Pylint back to 2.3.0. I manually upgraded it to 2.4.4 via `pip install --upgrade pylint`. However, putting `pylint_odoo` under `~/.pylintrc`’s `load-plugins` makes `pylint` produce the following error whenever it is ran:

```
Traceback (most recent call last):
  File "/home/devdesk4/.local/lib/python3.5/site-packages/pylint_plugin_utils/__init__.py", line 3, in <module>
    from pylint.utils import UnknownMessage
ImportError: cannot import name 'UnknownMessage'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/devdesk4/.local/bin/pylint", line 8, in <module>
    sys.exit(run_pylint())
  File "/home/devdesk4/.local/lib/python3.5/site-packages/pylint/__init__.py", line 23, in run_pylint
    PylintRun(sys.argv[1:])
  File "/home/devdesk4/.local/lib/python3.5/site-packages/pylint/lint.py", line 1686, in __init__
    linter.load_plugin_modules(plugins)
  File "/home/devdesk4/.local/lib/python3.5/site-packages/pylint/lint.py", line 660, in load_plugin_modules
    module = modutils.load_module_from_name(modname)
  File "/home/devdesk4/.local/lib/python3.5/site-packages/astroid/modutils.py", line 202, in load_module_from_name
    return load_module_from_modpath(dotted_name.split("."), path, use_sys)
  File "/home/devdesk4/.local/lib/python3.5/site-packages/astroid/modutils.py", line 245, in load_module_from_modpath
    module = imp.load_module(curname, mp_file, mp_filename, mp_desc)
  File "/usr/lib/python3.5/imp.py", line 244, in load_module
    return load_package(name, filename)
  File "/usr/lib/python3.5/imp.py", line 216, in load_package
    return _load(spec)
  File "<frozen importlib._bootstrap>", line 693, in _load
  File "<frozen importlib._bootstrap>", line 673, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 665, in exec_module
[...]

PR #285, opened 2020-06-24
[REF] requirements.txt: Update packages version

moylop260:
Closes https://github.com/OCA/pylint-odoo/issues/281

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,16 +1,12 @@
-astroid==1.6.5 ;python_version < '3'
-pylint==1.9.3 ;python_version < '3'
-pylint-plugin-utils==0.4 ;python_version < '3'
-
-astroid==2.2.0 ;python_version >= '3'
-pylint==2.3.0 ;python_version >= '3'
-pylint-plugin-utils==0.5 ;python_version >= '3'
-
-docutils==0.14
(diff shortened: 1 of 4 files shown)


requirements.txt
```diff
@@ -1,16 +1,12 @@
...
-
-astroid==2.2.0 ;python_version >= '3'
-pylint==2.3.0 ;python_version >= '3'
-pylint-plugin-utils==0.5 ;python_version >= '3'
-
-docutils==0.14
-lxml>=4.2.3
-Pygments==2.2
-restructuredtext_lint==1.1.3
+docutils==0.16
```
pedrobaeza, line 1: Not sure if this will collapse with MQT + ocabot version, as it's fixed in 0.15.1.

cc @sbidoul

sbidoul: That should be ok, since it was pinned at 0.14 before which is also different from 0.15.1.

pedrobaeza: Isn't the upper one used then (now 0.16)?

pedrobaeza: OK, not sure what is installed in each part. I was just rising the possible problem. It seems you are OK anyway, as README generator is handled outside Travis scope.

PR #286, opened 2020-07-04
[REF] missing-import-error: Add new packages for Odoo v13.0

moylop260:
Identified running:

 - `PYTHONPATH=~/pylint-odoo python3 -m pylint -d all -e missing-import-error,missing-manifest-dependency --load-plugins=pylint_odoo addons/* odoo/addons/*`

 - [PyPDF2](
https://github.com/odoo/odoo/blob/47707e1b099195487fd3d3e0ae012e4aed5a9bc5/requirements.txt#L33)
- [chardet](https://github.com/odoo/odoo/blob/47707e1b099195487fd3d3e0ae012e4aed5a9bc5/requirements.txt#L2)
- [urllib3 (requests dependency)](https://github.com/odoo/odoo/blob/47707e1b099195487fd3d3e0ae012e4aed5a9bc5/requirements.txt#L40)

### DIFF
--- a/pylint_odoo/checkers/modules_odoo.py
+++ b/pylint_odoo/checkers/modules_odoo.py
@@ -168,7 +168,7 @@ DFLT_IMPORT_NAME_WHITELIST = [
     # self-odoo
     'odoo', 'openerp',
     # Known external packages of odoo
-    'PIL', 'anybox.testing.openerp', 'argparse', 'babel',
+    'PIL', 'PyPDF2', 'anybox.testing.openerp', 'argparse', 'babel', 'chardet',
     'dateutil', 'decorator', 'docutils', 'faces', 'feedparser',
     'gdata', 'gevent', 'greenlet', 'jcconv', 'jinja2',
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Shouldn't it be conditional depending on the Odoo version? If not, importing that libraries in previous versions won't raise the warning.

moylop260:
Hi Pedro, you are right!

In fact, this check requires a full refactoring.
 - The point that you comment
 - Support the following case:
   - https://github.com/odoo/odoo/pull/25549

This PR is not making a full refactoring
It is only fixing a false red for new know libraries and it is blocking PR 🔴

If you are using that 3 libraries in <=Odoo-12.0 the test's build will be red if you didn't add the package to `requirements.txt` in in the meantime, it is a reviewer's warning to request the manifest dependency.

We can create an issue in order to have it in the radar.

pedrobaeza:
As you prefer. It can be better to avoid false reds instead of having false greens.

moylop260:
Hi Pedro
So 
Can we merge it?

Issue #219, opened 2018-10-27
E8103 sql-injection


CGenie:
Hello,
For the `sql-injection` checker you do check for `execute` and `executemany` methods and disallow variable creation with `%`, `+`, and `format`. However this is a valid way to create an SQL as described in the psycopg2 docs (https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#no-sql-injection):
```
>>> cur.execute(                                                # correct
...     SQL("INSERT INTO {} VALUES (%s)").format(Identifier('numbers')),
...     (10,))
```
Notice that the above code contains `execute` and `format` yet is correct and protected against SQL injections. I guess you should check additionally whether the variable is wrapped in `SQL` and `Identifier` and such are used in `format` variables.
I'm currently having an issue with the checker at https://github.com/OCA/hr/pull/478 so it's not a made-up example.
I'm also wondering how did this code
https://github.com/OCA/hr/blob/11.0/hr_holidays_hour/report/hr_holidays_report.py#L50-L52
pass through.

moylop260:
Thanks for identifying another case of this important check

CGenie:
Yeah, for now I'm trying to skip the check via this hack https://github.com/OCA/hr/pull/478/files#diff-e776bfa9f5e992680d0bd15f7ef46facR63 but I guess after you fix this, that can be removed as well.

pedrobaeza:
@moylop260 this has arisen again, can you please check for fixing it?

moylop260:
Hi @pedrobaeza 
Where was it raised again (in order to reproduce it locally)?

pedrobaeza:
https://travis-ci.com/github/OCA/server-tools/jobs/358747391#L381

for https://github.com/OCA/server-tools/pull/1858

moylop260:
Please, check the following fix:
https://github.com/OCA/pylint-odoo/pull/287

PR #287, opened 2020-07-11
[FIX] sql-injection: Consider valid arguments built with psycopg2.sql

moylop260:
Fix https://github.com/OCA/pylint-odoo/issues/219


yvaucher:
Yay! thanks for that @moylop260

PR #288, opened 2020-07-11
[IMP] Add openupgradelib as a whitelist.

feketemihai:
Lately i had some travis fails due to openupgradelib is not in the external dependencies, i don't know if we should add it to all the modules that contains migration or we can add it separately in the repositories, in isort.cfg
https://github.com/OCA/pylint-odoo/issues/247
or will work with the changes i did.

https://travis-ci.org/github/OCA/commission/jobs/707123086


feketemihai:
@moylop260 @pedrobaeza seems that this is not enough, still the builds are failing, do you how we can have general fix, not adding openupgradelib as dependency in all modules with migration?

moylop260:
A new version of pylint was released:
 - https://pypi.org/project/pylint-odoo/#changes

You will need update the pinned version to `3.3.2`:
 - https://github.com/OCA/commission/blob/c87f628913849f6b7854eefc8e20ad71890bf3f6/.pre-commit-config.yaml#L84-L88

PR #301, opened 2020-12-06
[REF] development-status-allowed: Explicitly report possible values

dreispt:


moylop260 (approved):
👍 
Thank you!

I think next case should be get the value from pylint.conf file
Since that it is a default but not the final one if you are using a custom one

dreispt:
@moylop260 Tests are failing, I'm not sure why. Can you assist please?

moylop260:
@dreispt 

Done!

The issue was a comma `,`

And I use the dynamic values from configuration file in order to set the correct values if you have a custom configuration file

sbidoul (approved):

moylop260:
@dreispt 

Could you check my last changes and give us your review, please?

In order to merge it

Regards!

Issue #302, opened 2020-12-07
New check: .format() over translated string _()


Yajo:
Today I noticed in https://github.com/OCA/server-tools/pull/1941#discussion_r537426404 that our friend @legalsylvain didn't know about the implicit dangers of using [format string syntax](https://docs.python.org/3/library/string.html#formatstrings) for translated content. Certainly it's not something very obvious, but [it's quite dangerous](https://lucumr.pocoo.org/2016/12/29/careful-with-str-format/), as it lets a translator to inject malicious translations that execute server-side code in the odoo server. Translated content must always use [printf style](https://docs.python.org/3/library/stdtypes.html#printf-style-string-formatting) formatting, which is dumber but secure.

Until today, I have never noticed any such usage in our code base, but I think a linter would be very helpful here.

pedrobaeza:
Not only dangerous, but can't be possible to be used with translatable fields, as the symbol replacement is done on linking time, and thus, with that, you won't have any translation.

Yajo:
I think you're talking about [f-strings](https://docs.python.org/3/reference/lexical_analysis.html#f-strings), which is a different thing (that indeed could also be linted)

AFAIK yes you could have a translation with `_("My name is {}").format("Jairo")`

legalsylvain:
@Yajo thanks a lot !

pedrobaeza:
Yes, I'm taking about f-strings

moylop260:
@Yajo 
So, Should  we do a dance to kukulcan to resurrect the following old proposal?
 - https://github.com/OCA/pylint-odoo/pull/65

moylop260:
Could you check the following PR, please?
 - https://github.com/OCA/pylint-odoo/pull/304

moylop260:
@Yajo 
Could you make a PR to change the following section:
 - [Prefer % over .format(), prefer %(varname) instead of positional. This is better for translation and clarity.](https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#33idioms)

In order to warn a security risk instead of a translation matter

Please

Yajo:
>  So, Should we do a dance to kukulcan to resurrect the following old proposal?

No, that one's fixed by python 3 which is the norm these days (finally). The problem we're talking here is just about translations.

Yajo:
>  Could you make a PR to change the following section:
> 
[...]

PR #304, opened 2020-12-07
[ADD] str-format-used: Check if str.format was used

moylop260:
More info: https://lucumr.pocoo.org/2016/12/29/careful-with-str-format/

Closes https://github.com/OCA/pylint-odoo/issues/302

### DIFF
--- a/pylint_odoo/checkers/no_modules.py
+++ b/pylint_odoo/checkers/no_modules.py
@@ -137,6 +137,13 @@ ODOO_MSGS = {
         'manifest-maintainers-list',
         settings.DESC_DFLT
     ),
+    'E%d05' % settings.BASE_NOMODULE_ID: (
+        'Use of `str.format` method in a translated string. '
+        'Use `"%(varname)s" % {"varname": value}` instead. '
+        'Be careful https://lucumr.pocoo.org/2016/12/29/careful-with-str-format',
+        'str-format-used',
+        settings.DESC_DFLT
+    ),
     'C%d01' % settings.BASE_NOMODULE_ID: (
         'One of the following authors must be present in manifest: %s',
         'manifest-required-author',
@@ -485,6 +492,7 @@ class NoModuleChecker(misc.PylintOdooChecker):
                           'translation-required',
                           'translation-contains-variable',
                           'print-used',
+                          'str-format-used',
                           )
     def visit_call(self, node):
         infer_node = utils.safe_infer(node.func)
@@ -605,6 +613,7 @@ class NoModuleChecker(misc.PylintOdooChecker):
                     and isinstance(arg.func, astroid.Attribute)
                     and isinstance(arg.func.expr, astroid.Const)
                     and arg.func.attrname == 'format'):
+                self.add_message('str-format-used', node=node)
                 wrong = arg.as_string()
                 params_as_string = ', '.join([
                     x.as_string()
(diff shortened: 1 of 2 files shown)


luisg123v (approved):
:+1:

pylint_odoo/checkers/no_modules.py
```diff
@@ -226,6 +226,12 @@
         'print-used',
         settings.DESC_DFLT
     ),
+    'W%d17' % settings.BASE_NOMODULE_ID: (
```
Yajo, line 229: I would consider this an error, not a warning. It introduces security issues.

pylint_odoo/checkers/no_modules.py
```diff
@@ -479,11 +485,15 @@ def visit_print(self, node):
...
                           'print-used',
+                          'str-format-used',
                           )
     def visit_call(self, node):
         infer_node = utils.safe_infer(node.func)
-        if utils.is_builtin_object(infer_node) and infer_node.name == 'print':
-            self.add_message('print-used', node=node)
+        if utils.is_builtin_object(infer_node):
+            if self.get_func_name(node.func) == 'format':
+                self.add_message('str-format-used', node=node)
```
Yajo, line 494: Sorry because I don't understand fine this code, but are you checking the source of the string?

I mean, using `"This is a {}".format(thing)` is :+1:, but  `_("This is a {}").format(thing)` is :-1:. I would only forbid `.format()` when the pattern string is untrusted.

moylop260: This code is checking if `*.format` is used.

I didn't get you.

> I mean, using `"This is a {}".format(thing)` is :+1:, but  `_("This is a {}").format(thing)` is :-1:. I would only forbid `.format()` when the pattern string is untrusted.

I didn't get you.

Why `"This is a {}".format(thing)` is good?

moylop260: You can test locally this branch using the following steps:

```bash
git clone https://github.com/vauxoo-dev/pylint-odoo.git -b master-oca-avoid-format-moy /tmp/pylint-odoo
cd /tmp/pylint-odoo
python3 -m pip install -r requirements.txt
python3 -m pylint -d all -e str-format-used --load-plugins=pylint_odoo pylint_odoo/test_repo/
```

So, you can change the files in test_repo/* to test different cases

Yajo: > Why `"This is a {}".format(thing)` is good?

It is good because the template (`"This is a {}"`) is trusted.

If the template came from an untrusted source, such as a translation file or an argument in a public method, it would be dangerous.

fernandahf: Hi @Yajo 

I have made some changes considering you mentioned.

Could you review again, please?

Regards.

Yajo (approved):

pylint_odoo/checkers/no_modules.py
```diff
@@ -137,6 +137,13 @@
         'manifest-maintainers-list',
         settings.DESC_DFLT
     ),
+    'E%d05' % settings.BASE_NOMODULE_ID: (
+        'Use of `str.format` method in a translated string. '
+        'Use `"%(varname)s" % {"varname": value}` instead. '
```
moylop260, line 142:
```suggestion
        'Use `_("%(varname)s") % {"varname": value}` instead. '
```

Issue #303, opened 2020-12-07
New check: translated strings containing unordered placeholders


Yajo:
As one can see in https://github.com/odoo/odoo/issues/29104 (not related to this proposal, but gives an example), there are some sentences which are hard to translate from English to other languages if you can't alter the order of words.

So, my proposal is that we add a linter that ensures that, when a translated string has more than 1 interpolation, it must be expanded with a dict and not a tuple.

Example:

```py
_("Your %s %s is shipping") % ("red", "car")
```

It seems nice in English, but in Spanish you would have to use an unnatural sentence such as `"Su envío, que es de color %s y es un/a %s, está de camino"`.

The original code should have been:

```py
_("Your %(color)s %(item)s is shipping") % {"color": "red", "item": "car"}
```

So you can translate that into `"Su %(item)s %(color)s está enviándose"` in Spanish.

If there's just one `%s` or the string is not translatable, this is not an issue.

FernandoRomera:
IIUC v14.0 supports passing arguments to translation lookup.
https://www.odoo.com/documentation/14.0/reference/translations.html#variables

before 14
_("Your %s %s is shipping") % ("red", "car")

now is
_("Your %s %s is shipping", "red", "car")

pedrobaeza:
But the order question remains the same.

FernandoRomera:
My note is for the PR #305, if it can adjust positional string formatting or passing arguments?

Test ok if:
_("Your %(color)s %(product)s is shipping", color="red", product="car")

and ko
_("Your %s %s is shipping", "red", "car")

Yajo:
Good, I didn't know of that new feature. Indeed we should fail if it's called with more than 1 `*args` and ask to use `**kwargs` instead.

PR #305, opened 2020-12-07
[ADD] translation-positional-used: New check to valid the allow change the order of string formatting

moylop260:
If you are using:
 - `_("Your %s %s is shipping") % ("red", "car")`

Better using:
 - `_("Your %(color)s %(item)s is shipping") % {"color": "red", "item": "car"}`

In order to allow other languages change the order
 - e.g. in Spanish: `"Su %(item)s %(color)s está enviándose"`

Fix https://github.com/OCA/pylint-odoo/issues/303

### DIFF
--- a/pylint_odoo/checkers/no_modules.py
+++ b/pylint_odoo/checkers/no_modules.py
@@ -232,6 +232,13 @@ ODOO_MSGS = {
         'print-used',
         settings.DESC_DFLT
     ),
+    'W%d20' % settings.BASE_NOMODULE_ID: (
+        'Translation method _(%s) is using positional string formatting. '
+        'Use named placeholder string formatting "_(%%(placeholder)s)" '
+        'or "_({placeholder})" instead.',
+        'translation-positional-used',
+        settings.DESC_DFLT
+    ),
     'F%d01' % settings.BASE_NOMODULE_ID: (
         'File "%s": "%s" not found.',
         'resource-not-exist',
@@ -484,7 +491,7 @@ class NoModuleChecker(misc.PylintOdooChecker):
                           'renamed-field-parameter',
                           'translation-required',
                           'translation-contains-variable',
-                          'print-used',
+                          'print-used', 'translation-positional-used',
                           )
     def visit_call(self, node):
         infer_node = utils.safe_infer(node.func)
@@ -593,6 +600,7 @@ class NoModuleChecker(misc.PylintOdooChecker):
                 and node.func.name == '_'
                 and node.args):
             wrong = ''
+            right = ''
             arg = node.args[0]
             # case: _('...' % (variables))
             if isinstance(arg, astroid.BinOp) and arg.op == '%':
@@ -611,10 +619,32 @@ class NoModuleChecker(misc.PylintOdooChecker):
                     for x in itertools.chain(arg.args, arg.keywords or [])])
                 right = '_(%s).format(%s)' % (
                     arg.func.expr.as_string(), params_as_string)
-            if wrong:
+            if wrong and right:
                 self.add_message(
                     'translation-contains-variable', node=node,
                     args=(wrong, right))
+
+            # translation-positional-used: Check "string to translate"
+            # to check %s or {} used
+            str2translate = arg.as_string()
+            extra_data = {}
+            msgid_args, msgid_kwargs = (
+                misc.WrapperModuleChecker.
+                _get_format_str_args_kwargs(str2translate, extra_data=extra_data))
+            is_positional = False
+            if extra_data['has_placeholders']:
+                if len(msgid_args) >= 2 and extra_data['has_positional']:
+                    is_positional = True
+            else:
+                printf_args = (
+                    misc.WrapperModuleChecker.
+                    _get_printf_str_args_kwargs(str2translate))
+                # Return tuple for %s and dict for %(varname)s
+                if isinstance(printf_args, tuple) and len(printf_args) >= 2:
+                    is_positional = True
+            if is_positional:
+                self.add_message('translation-positional-used',
+                                 node=node, args=(str2translate,))
 
         # SQL Injection
         if isinstance(node, astroid.Call) and node.args and \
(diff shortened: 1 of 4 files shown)


Yajo:
There's something wrong here. You're also checking strings such as `"Your {} {} is shipping"`. But that's not a problem because:

1. If you're using format string syntax, you would be able to use `"Su {1} {0} está enviándose"`.
1. That syntax should be instead forbidden completely due to security problems as explained in #304.

moylop260:
@Yajo 
Good point!

Could you check it again, please?

I just considered the `printf` cases
The format ones could be changed as you said but avoid from the another lint-check (no this one).

Thanks!

pylint_odoo/checkers/no_modules.py
```diff
@@ -605,11 +612,23 @@ def visit_call(self, node):
...
                 self.add_message(
                     'translation-contains-variable', node=node,
                     args=(wrong, right))
 
+            # translation-positional-used: Check "string to translate"
+            # to check "%s %s..." used where the position can't be changed
+            str2translate = arg.as_string()
+            printf_args = (
+                misc.WrapperModuleChecker.
+                _get_printf_str_args_kwargs(str2translate))
```
Yajo, line 625:
```suggestion
                misc._get_printf_str_args_kwargs(str2translate))
```

pylint_odoo/checkers/no_modules.py
```diff
@@ -226,6 +226,12 @@
         'print-used',
         settings.DESC_DFLT
     ),
+    'W%d20' % settings.BASE_NOMODULE_ID: (
+        'Translation method _(%s) is using positional string printf formatting. '
+        'Use named placeholder "_(%%(placeholder)s)" instead.',
```
Yajo, line 231:
```suggestion
        'Use named placeholder `_("%%(placeholder)s")` instead.',
```

I guess the double-percent symbol is needed, right?

pylint_odoo/test_repo/broken_module/models/broken_model.py
```diff
@@ -245,6 +245,23 @@ def my_method1(self, variable1):
         error_msg = _('Variable not translatable: {}'.format(variable1))
         error_msg = _('Variables not translatable: {}, {variable2}'.format(
             variable1, variable2=variable2))
+
+        # string with parameters without name
+        # so you can't change the order in the translation
+        _('%s %d') % ('hello', 3)
+        _('%s %s') % ('hello', 'world')
+        _('%s %s %s') % ('hello', 'world!')
```
Yajo, line 253: This seems like a different failure. I'm not sure if normal pylint would detect it. You're using 3 paceholders with 2 args.

Possibly not interesting to implement, given in this case you should use named placeholders.

pylint_odoo/test_repo/broken_module/models/broken_model.py
```diff
@@ -245,6 +245,23 @@ def my_method1(self, variable1):
...
+        # Valid cases
+        _('{0} {1}') % ('hello', 3)
+        _('%(strname)s') % {'strname': 'hello'}
+        _('%(strname)s %(intname)d') % {'strname': 'hello', 'intname': 3}
+        _('%s') % 'hello'
+        _('%d') % 3
+        _('{}') % 'hello'
+        _('{:d}') % 3
+        _('{} {}') % ('hello', 3)  # It can be translated to '{1} {0}'
+        _('{} {:d}') % ('hello', 3)
```
Yajo, line 264: Are these tests worthwhile?

pylint_odoo/test_repo/broken_module/models/broken_model.py
```diff
@@ -245,6 +245,23 @@ def my_method1(self, variable1):
...
             variable1, variable2=variable2))
+
+        # string with parameters without name
+        # so you can't change the order in the translation
+        _('%s %d') % ('hello', 3)
+        _('%s %s') % ('hello', 'world')
+        _('%s %s %s') % ('hello', 'world!')
+
+        # Valid cases
+        _('{0} {1}') % ('hello', 3)
```
Yajo, line 256: Is it worth it?

pylint_odoo/checkers/no_modules.py
fernandahf: Done

pylint_odoo/checkers/no_modules.py
fernandahf: Hi @Yajo 

This is full line:

```
misc.WrapperModuleChecker._get_printf_str_args_kwargs(str2translate))
```
`_get_printf_str_args_kwargs` comes from `WrapperModuleChecker` class.

pylint_odoo/test_repo/broken_module/models/broken_model.py
fernandahf: I think that is a different failure. I removed it.

pylint_odoo/test_repo/broken_module/models/broken_model.py
fernandahf: I think a few examples is enough.

pylint_odoo/test_repo/broken_module/models/broken_model.py
fernandahf: I think a few examples is enough. I removed a few examples.

pylint_odoo/checkers/no_modules.py
Yajo: Oh sorry, the formatting tricked me. 🤦🏼‍♂️

pylint_odoo/test_repo/broken_module/models/broken_model.py
Yajo: This line is not a valid case. It uses str-format syntax with `%`

fernandahf: You right, I removed it.

Yajo (approved):

fernandahf:
@luisg123v 

Regards.

PR #307, opened 2020-12-16
[12.0][REF] pylint-odoo: odoo-addons-relative-import does not apply in test dir

### LINKED ISSUE #297: odoo-addons-relative-import should not apply in tests/*
This is a subject I raised several times on some PRs (last one here https://github.com/akretion/connector/pull/2#discussion_r509111937) but never opened an issue until now...

Currently, if I develop the module `my_module`, if I write a test with:

```
from odoo.addons.my_module.foo import bar
```

The `odoo-addons-relative-import` warning asks to write:

```
from ..foo import bar
```

I agree that relative imports must be used in the regular module files (models, controllers, ...). But IMO it is more correct to use absolute imports in tests. Tests are not truly constitutive of the module, they just happen to be stored inside the module. Usually, in Python libs, they're stored outside of the package:

```
 requests $ tree -L 1       
.
├── AUTHORS.rst
├── docs
├── ...
├── requests
├── requirements.txt
├── setup.py
└── tests
```

Importantly, using the absolute import proves that the addon we develop works *as a library* too.

Assume this example (an issue we actually encountered, [e.g.](https://github.com/OCA/server-env/blob/80a56d320a684b0e2eb22c9daaeab56a323d19e0/server_environment/__init__.py#L21-L25)):

```
.
├── my_module
    ├── __init__.py
    ├── foo.py
    └── tests
        ├── __init__.py
        └── test_my_module.py
```

In `foo.py`, we have:

```
def foo():
    # do stuff

def bar():
    # do other stuff
```

And in `my_module/__init__.py`:

```
from .foo import foo
```

If in your tests, you want to test the methods `foo` and `bar` and use relative imports:

```
from ..foo import foo, bar

def test_foo():
     # ...

def test_bar():
    # ...
```

You'll not realize that `bar` cannot be imported at all from other addons.
But if you test using absolute imports:

```
from odoo.addons.my_module.foo import foo, bar  # fails
```

You'll realize soon enough that you broke the imports, because the module `foo` is shadowed by the method `foo`.

fernandahf:

It is more correct to use absolute imports in tests. Tests are not truly constitutive of the module, they just happen to be stored inside the module. Usually, in Python libs, they're stored outside of the package. Importantly, using the absolute import proves that the addon we develop works as a library too.

Current behavior before PR: pylint-odoo warns about relative imports in tests/*

Desired behavior after PR is merged: pylint-odoo does not warn about relative imports in tests/*

Fix #297 

-- I confirm I have signed the CLA and read the PR guidelines at www.odoo.com/submit-pr

### DIFF
--- a/pylint_odoo/checkers/modules_odoo.py
+++ b/pylint_odoo/checkers/modules_odoo.py
@@ -325,6 +325,15 @@ class ModuleChecker(misc.WrapperModuleChecker):
 
     def _get_odoo_module_imported(self, node):
         odoo_module = []
+        if self.manifest_file and hasattr(node.parent, 'file'):
+            relpath = os.path.relpath(
+                node.parent.file, os.path.dirname(self.manifest_file))
+            if os.path.dirname(relpath) == 'tests':
+                # import errors rules don't apply to the test files
+                # since these files are loaded only when running tests
+                # and in such a case your
+                # module and their external dependencies are installed.
+                return odoo_module
         if isinstance(node, astroid.ImportFrom) and \
                 ('openerp.addons' in node.modname or
                  'odoo.addons' in node.modname):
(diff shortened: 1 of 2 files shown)


fernandahf:
@moylop260 

PR #315, opened 2021-01-19
[FIX] test-folder-imported: check if node has a modname as attribute

fernandahf:

In python versions above 3.5, following error occurs when try to get `modname` from node.

`AttributeError: 'Import' object has no attribute 'modname'`

Current behavior before PR:
Error caused by try to get `modname` from node from `Import` object.

Desired behavior after PR is merged:
No errors shown when try to get `modname` from node.

Fix https://github.com/OCA/pylint-odoo/issues/318

### DIFF
--- a/pylint_odoo/checkers/modules_odoo.py
+++ b/pylint_odoo/checkers/modules_odoo.py
@@ -366,11 +366,15 @@ class ModuleChecker(misc.WrapperModuleChecker):
 
     def check_folder_test_imported(self, node):
         if (hasattr(node.parent, 'file')
-            and os.path.basename(node.parent.file) == '__init__.py'
-            and ("tests" in [item for item, _ in node.names]
-                 or "tests" in node.modname)):
-            self.add_message('test-folder-imported', node=node,
-                             args=(node.parent.name,))
+                and os.path.basename(node.parent.file) == '__init__.py'):
+            package_names = []
+            if isinstance(node, astroid.ImportFrom):
+                package_names = node.modname.split('.')[:1]
+            elif isinstance(node, astroid.Import):
+                package_names = [name[0].split('.')[0] for name in node.names]
+            if "tests" in package_names:
+                self.add_message('test-folder-imported', node=node,
+                                 args=(node.parent.name,))
 
     @staticmethod
     def _is_absolute_import(node, name):


pylint_odoo/checkers/modules_odoo.py
```diff
@@ -368,7 +368,8 @@ def check_folder_test_imported(self, node):
         if (hasattr(node.parent, 'file')
             and os.path.basename(node.parent.file) == '__init__.py'
             and ("tests" in [item for item, _ in node.names]
-                 or "tests" in node.modname)):
+                 or (hasattr(node, 'modname')
+                    and "tests" in node.modname))):
```
luisg123v, line 372: Pro tip:  
Instead of:  
`hasattr(node, 'modname and 'test' in node.modname`  

It may be simplified as:  
`'test' in getattr(node, 'modename', '')`

Similar as dictionary, you may use a getter with a default method.

luisg123v (changes requested):
Title: "like" -> "as".

"Like" is used to compare things or to indicate they're similar, while "as" may be used to indicate something is used in place of, or fulfilling the role of.

pylint_odoo/checkers/modules_odoo.py
moylop260: A import node could be a `import package` or `from . import package`

It has 2 different classes:
 - astroid.ImportFrom
 - astroid.Import

You are just validating the case:
  - `from . import tests`

So, you can use the following validation:
 - https://github.com/OCA/pylint-odoo/blob/b4dae33082192c7a0b85a19cf9e7c68e35516581/pylint_odoo/checkers/no_modules.py#L448-L449

Or if you want to support the following case:
  - `import tests`

You can se the following validation:
 - https://github.com/OCA/pylint-odoo/blob/b4dae33082192c7a0b85a19cf9e7c68e35516581/pylint_odoo/checkers/no_modules.py#L450-L451

This code will be compatible with multi-imports lines like
  - `from . import tests, models`
  -  `import tests, models`
 
Regargs!

pylint_odoo/checkers/modules_odoo.py
```diff
@@ -366,11 +366,14 @@ def check_odoo_relative_import(self, node):
 
     def check_folder_test_imported(self, node):
         if (hasattr(node.parent, 'file')
-            and os.path.basename(node.parent.file) == '__init__.py'
-            and ("tests" in [item for item, _ in node.names]
-                 or "tests" in node.modname)):
-            self.add_message('test-folder-imported', node=node,
-                             args=(node.parent.name,))
+                and os.path.basename(node.parent.file) == '__init__.py'):
+            if isinstance(node, astroid.ImportFrom):
```
moylop260, line 370: What about initialize the variable `package_names = []`

In order to avoid the following error: `NameError: name 'package_names' is not defined` if the method is called using a node different to `*Import` classes

```suggestion
                and os.path.basename(node.parent.file) == '__init__.py'):
            package_names = []
            if isinstance(node, astroid.ImportFrom):
```

fernandahf: Done

PR #320, opened 2021-01-20
[REF] requirements: upgrade pylint version

### LINKED ISSUE #319: Upgrade pylint==2.5.3 to pylint==2.6.0 for py>=3
The following line of code:
 - https://github.com/OCA/pylint-odoo/blob/1b9018e3884fceced0a06a1a19b29ffaffa5f701/requirements.txt#L8

is pinning the pylint version because it could raise red with newer versions but we need keep upgrading this version each period of time

There is a new release pylint-2.6.0 on Aug 21, 2020 

The change log is the following one:
 - https://github.com/PyCQA/pylint/blob/2.6/ChangeLog#L5-L64

It is required to upgrade in order to have these fixes and new checkers even if they are not activated from configuration file

Note: For py<3 there is not available new versions since that py2.x is deprecated now

fernandahf:

The following line of code:

- https://github.com/OCA/pylint-odoo/blob/1b9018e3884fceced0a06a1a19b29ffaffa5f701/requirements.txt#L8

is pinning the pylint version because it could raise red with newer versions but we need keep upgrading this version each period of time

There is a new release pylint-2.6.0 on Aug 21, 2020

Current behavior before PR:
pylint-odoo works correctly.

Desired behavior after PR is merged:
pylint-odoo works correctly with new pylint version.

Fix #319 

-- I confirm I have signed the CLA and read the PR guidelines at www.odoo.com/submit-pr

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,11 +1,12 @@
 docutils==0.16
+isort==4.3.21
 lxml>=4.2.3
 polib==1.1.0
 Pygments==2.2 ;python_version < '3'
 Pygments==2.6.1 ;python_version >= '3'
 pylint-plugin-utils==0.6
 pylint==1.9.5 ;python_version < '3'
-pylint==2.5.3 ;python_version >= '3'
+pylint==2.6.0 ;python_version >= '3'
 restructuredtext_lint==1.3.1
 rfc3986
 six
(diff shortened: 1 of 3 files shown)


fernandahf:
@moylop260 

Regards.

requirements.txt
```diff
@@ -1,11 +1,12 @@
 docutils==0.16
+isort==4.3.21
```
moylop260, line 2: Better check how we can avoid this pinned version, please

fernandahf: Could you review again, please?

Regards.

PR #325, opened 2021-04-09
Add .pylintrc example

tmotyl:
Pylint-odoo shoud come with example .pylintrc file one can copy to project and start using right away.

This file is taken from https://github.com/OCA/server-env/blob/13.0/.pylintrc

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -157,9 +157,9 @@ Usage
 
 ``pylint --load-plugins=pylint_odoo -e odoolint path/to/test``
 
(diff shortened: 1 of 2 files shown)


tmotyl:
not sure whether the `-e odoolint` param is needed, as the https://github.com/OCA/oca-addons-repo-template/blob/master/.pre-commit-config.yaml.jinja doesn't have it.

jgrandguillaume (approved):
LGTM, thx for the contribution.

PR #328, opened 2021-04-13
[FIX] Add pbr to requirements

joao-p-marques:
The library is needed for the setup, but it is not specified in requirements.
Its installation depends on setuptools' `easy_install`, which is unreliable: 
- https://docs.openstack.org/pbr/latest/#pbr-python-build-reasonableness 
- https://pip.pypa.io/en/latest/cli/pip_install/#controlling-setup-requires

Example of the error that might be caused: https://github.com/Tecnativa/doodba/runs/2314862208?check_suite_focus=true#step:7:1476
Pre-installing `pbr` with pip solves the issue, so it could be added to the `requirements.txt` to force the installation that way.

ping @Yajo

Issue #306, opened 2020-12-10
Dependencies break with pip 20.3


joao-p-marques:
`pip` 20.3 introduced a new dependency resolver that disallows having conflicting dependencies in the same environment. https://pip.pypa.io/en/stable/user_guide/?highlight=resolver#changes-to-the-pip-dependency-resolver-in-20-3-2020

As `pylint-odoo` requires `pylint==2.5.3` (for python 3), a dependency conflict is easily obtained when trying to install it along with `pylint`:
```
conflict caused by:
        The user requested isort==4.3.21
        pylint 2.6.0 depends on isort<6 and >=4.2.5
        pylint-odoo 3.5.0 depends on isort==4.3.4
```
and
```
        The user requested pylint 2.6.0
        pylint-odoo 3.5.0 depends on pylint==2.5.3
```

This forces the use of an older `pylint` and `isort` version, which is not ideal. 
What is the reason for requesting that specific version? Can the requirement be more flexible (like `>=` or similar)?
My suggestion would be to support the same `isort` constraint as the one used by the required `pylint` version.

cc @Tecnativa
TT27232

ping @Yajo

PR #329, opened 2021-04-13
[IMP] Migrate to poetry to manage package

joao-p-marques:
The fix from https://github.com/OCA/pylint-odoo/pull/328 is incomplete. 
The problem still persists and will remain (and possibly escalate) while using deprecated packaging methods.

As mentioned in the following references:
- https://pip.pypa.io/en/latest/cli/pip_install/#controlling-setup-requires
- https://setuptools.readthedocs.io/en/latest/userguide/dependency_management.html?highlight=dependencies#id3
- pypa/setuptools#1320 (comment)

It is recommended to update the packaging method not to depend on things like the `setup_requires` keyword, but instead manage the build dependencies from a `pyproject.toml` file.

To avoid complications when migrating to a different Python packaging method, [poetry](https://python-poetry.org) can facilitate the move.
It can fully replace `setup.py` and `setup.cfg`, while still integrating with current tools like `uniitest` and `tox`.
With this PR, I try to make that transition, while keeping the current functionality. 

The only current functionality I didn't manage to incorporate here was the automatic changelog generation from git. If it is an important thing to keep, it can be done with most up-to-date tools based on https://www.conventionalcommits.org/en/v1.0.0/, like https://commitizen-tools.github.io/commitizen/ to read from git history and generate the changelog, like now. It would need some configuration to adapt it to the current OCA commit guidelines, but it could be done and would allow a bunch of cool features like https://commitizen-tools.github.io/commitizen/#commitizen-features
WDYT @moylop260?

@Tecnativa
TT29203

ping @Yajo


.travis.yml
```diff
@@ -8,44 +8,51 @@ cache:
...
+        PY=3.7
         DEPLOY="1"
     - python: 3.8
       env:
         TOXENV=pylint
+        PY=3.8
     - python: 3.9-dev
       env:
         TOXENV=pylint
+        PY=3.9-dev
```
Yajo, line 32: Python 3.9 is stable these days.

```suggestion
        PY=3.9
```

pyproject.toml
```diff
@@ -0,0 +1,49 @@
...
+whichcraft = "^0.6.1"
+pylint = [
+    {version = "^1.9.5", python = "^2.7"},
+    {version = "^2.6.0", python = "^3"}
+]
+Pygments = [
+    {version = "^2.2", python = "^2.7"},
+    {version = "^2.6.1", python = "^3"}
+]
+pylint-plugin-utils = "0.6"
```
Yajo, line 36: Usually, when you're building a library (or a plugin, like here), it's better not to be so specific about requirements, if possible. This avoids conflicts such as #306.
```suggestion
python = ">=2.7,<3.0.0 || >=3.5.0"
docutils = "^0.16"
lxml = "^4.6.3"
polib = "^1.1.0"
restructuredtext-lint = "^1.3.1"
rfc3986 = "^1.4.0"
six = "^1.15.0"
whichcraft = "^0.6.1"
pylint = [
    {version = "^1.9.5", python = "^2.7"},
    {version = "^2.6.0", python = "^3"}
]
Pygments = [
    {version = "^2.2", python = "^2.7"},
    {version = "^2.6.1", python = "^3"}
]
pylint-plugin-utils = "^0.6"
```

pyproject.toml
```diff
@@ -0,0 +1,49 @@
...
+
+[tool.poetry.dev-dependencies]
+nodeenv = "^1.6.0"
+coveralls = [
+    {version = "<2.0.0", python = "^2.7"},
+    {version = "^3.0.1", python = "^3"}
+]
+flake8 = "3.8.3"
```
Yajo, line 44:
```suggestion
flake8 = "^3.8.3"
```

Yajo (approved):
In general terms, poetry is pure pleasure.

Besides, it fixes what #328 was trying to fix (and didn't).

You have my blessing.

joao-p-marques:
@Yajo changes made. Thanks!

Yajo:
Please could you review @moylop260?

poetry.lock
```diff
@@ -0,0 +1,1013 @@
+[[package]]
```
moylop260, line 1: How did you make this file?

What extra considerations should we take now with this new way for each release?

.travis.yml
```diff
@@ -8,44 +8,51 @@ cache:
...
 script:
-    - flake8 --ignore=E722,F601,F841,W503,W504,W605 --max-line-length=88 --exclude=__init__.py .
-    - restructuredtext-lint ${TRAVIS_BUILD_DIR}/README.rst
     - tox -e $TOXENV,profile-stats
+    - poetry run flake8 --ignore=E722,F601,F841,W503,W504,W605 --max-line-length=88 --exclude=__init__.py ./pylint_odoo
+    - poetry run restructuredtext-lint ${TRAVIS_BUILD_DIR}/README.rst
 
 after_success:
-    - coveralls
-    - python setup.py sdist  # Build ChangeLog file from git log
```
moylop260, line 48: Where is building the package to upload to pypi?

poetry.lock
joao-p-marques: This file is managed by [poetry](https://python-poetry.org/docs/basic-usage/#installing-without-poetrylock). This is the file poetry uses to resolve all the dependencies for the environment and pin them to a specific version.

The only thing to take into account in a new release would be bumping the version in the `pyproject.toml` file. However, I think you can even enable another plugin in poetry to enable this automatically from git tags. @Yajo knows this better :smile:

.travis.yml
joao-p-marques: Oh sorry. I think I saw your comment there and assumed that step only enabled the changelog generation from git, so I deleted it.

We can build and push with [`poetry publish`](https://python-poetry.org/docs/cli/#publish)

I will be adding this.

joao-p-marques: We can either use `poetry build` to build the package and use the current deploy stage to push it, or use `poetry publish --build` which will do both in one step. Which do you prefer @moylop260?

poetry.lock
Yajo: Yes, you can use https://github.com/mtkennerly/poetry-dynamic-versioning/

ap-wtioit:
This seems to break our doobda tests when running 11.0 and 12.0 with python 3.6 (instead of 3.5), it works fine when installing the pypi pylint-odoo version 3.7.1:
```
doodba INFO: Executing /qa/venv/bin/python --version
Python 3.6.13
INFO:root:Subtest execution: test_qa (tests.ScaffoldingCase) (DB_VERSION='11', ODOO_MINOR='11.0', PWD='/home/gitlab-runner/builds/DAvFejyg/0/odoo_tools/docker-odoo-base/tests/scaffoldings/settings', command=('/qa/venv/bin/python', '-c', 'import pylint_odoo'))
Creating settings-295419_odoo_run ... 
Creating settings-295419_odoo_run ... done
doodba INFO: Waiting until postgres is listening at postgresql...
doodba WARNING: Attempt to install unaccent in prod@postgresql failed with this message: psql: error: FATAL: database "prod" does not exist
doodba INFO: Linking all addons from /opt/odoo/custom/src/addons.yaml in /opt/odoo/auto/addons
doodba INFO: Generating /opt/odoo/auto/odoo.conf file. Overriding any existing...
doodba INFO: Merging found configuration files in /opt/odoo/auto/odoo.conf
doodba INFO: Executing /qa/venv/bin/python -c import pylint_odoo
  File "<string>", line 1, in <module>
  File "/qa/venv/lib/python3.6/site-packages/pylint_odoo/__init__.py", line 2, in <module>
    from . import checkers
  File "/qa/venv/lib/python3.6/site-packages/pylint_odoo/checkers/__init__.py", line 1, in <module>
    from . import modules_odoo
  File "/qa/venv/lib/python3.6/site-packages/pylint_odoo/checkers/modules_odoo.py", line 13, in <module>
    from .. import misc, settings
  File "/qa/venv/lib/python3.6/site-packages/pylint_odoo/misc.py", line 14, in <module>
    from pylint.utils import _basename_in_blacklist_re
ImportError: cannot import name '_basename_in_blacklist_re'
1
INFO:root:Subtest execution: test_qa (tests.ScaffoldingCase) (DB_VERSION='11', ODOO_MINOR='11.0', PWD='/home/gitlab-runner/builds/DAvFejyg/0/odoo_tools/docker-odoo-base/tests/scaffoldings/settings')
Stopping settings-295419_postgresql_1 ... 
Stopping settings-295419_postgresql_1 ... done
Removing settings-295419_postgresql_1 ... 
Removing settings-295419_postgresql_1 ... done
Removing network settings-295419_default
Removing volume settings-295419_db
Removing volume settings-295419_filestore
```

joao-p-marques:
> This seems to break our doobda tests when running 11.0 and 12.0 with python 3.6 (instead of 3.5), it works fine when installing the pypi pylint-odoo version 3.7.1

Thanks @ap-wtioit 
Indeed, for python > 3.5 it was installing a latest version of pylint, which was incompatible with this plugin. We hadn't noticed because we only added it in Doodba for v11 and v12 for now (which use Python 3.5) and the CI uses the poetry lock file to install the dependencies, which had the correct version pinned.
Now that I pinned the version in the `pyproject.toml` and also enabled the build in CI, it shouldn't cause any problems.

Yajo:
I pushed the fix, you can undo the pin if you prefer.

joao-p-marques:
> I pushed the fix, you can undo the pin if you prefer.

Ok :+1: Thanks!
Aren't there any other potential changes that can introduce an error by upgrading the version?
@moylop260 is there another reason to have that version pinned?

moylop260:
Sorry, I didn't get you

What pinned version?

joao-p-marques:
> What pinned version?

Currently, the `requirements.txt` file pins the version of `pylint` with a strict `==2.6.0`: https://github.com/OCA/pylint-odoo/blob/master/requirements.txt#L9
When I adapted to poetry, I added a different constraint of `^2.6.0`, which was making all the versions up to `2.7.0` eligible, and led to some problems that @Yajo solved in https://github.com/OCA/pylint-odoo/pull/329/commits/8a6915146fa441e326521fef2800730185598af2
Now, we can keep that constraint slightly more loose or, if there were other reasons to keep that strict version number, we can also do that with https://github.com/OCA/pylint-odoo/pull/329/commits/f049dd427394a99535d30c1945d557c97df2d574

moylop260:
@joao-p-marques 

Ok, I got it

The reason is that this project is used for all projects of the OCA
So, if there is a new change in pylint incompatible with pylint-odoo so all CI will be red

We liked avoid it using only pinned version and bumping version using a new PR in order to check all is fine with the CI or fix the compatibility matter.

So, you can bump version in this PR in order to check all is working well or fix it if it is not

joao-p-marques:
Ok, thanks @moylop260 

With f049dd427394a99535d30c1945d557c97df2d574, `poetry` limits the version of `pylint` to only patch updates, you can see it in the docs: https://python-poetry.org/docs/dependency-specification/#tilde-requirements

So, in this case, it would limit to versions `2.6.0` up to `2.6.2`. I’ll do some more testing to make sure it doesn’t introduce any problem and report back.

joao-p-marques:
Well, taking a better look at it, the only `2.6.*` release in Github is `2.6.2` (`2.6.1` must have been deleted) and the only change is a fix in the Astroid version:
image
https://github.com/PyCQA/pylint/releases/tag/pylint-2.6.2
A quick test in a docker container and everything seems to work...

About this https://github.com/OCA/pylint-odoo/pull/329#discussion_r617233061 which approach do you prefer to follow then?

moylop260:
Sorry, I'm not familiar neither poetry and `^` and `~` dependencies structure

I see this PR hard to maintain from me in the time

@guewen What do you think?

Yajo (approved):

joao-p-marques:
> Sorry, I'm not familiar neither poetry and ^ and ~ dependencies structure
I see this PR hard to maintain from me in the time

Sorry, `poetry` also allows to specify dependency version constraints in a more conventional way, like `pip`. However, it does provide some extra syntax to make it easier, you can find more here: https://python-poetry.org/docs/dependency-specification/#version-constraints

Overall, it is a change to make it easier to maintain IMHO, specially with the changes in PEP standards I mention above and that will only cause more problems over time. I understand this is a big change, so it's up to you if it is worth it to invest in a tool like this 🙂

moylop260:
@yajo @joao-p-marques 

I have created a new python packages here:
 - https://github.com/oca/odoo-pre-commit-hooks

I just deleted the requirement to `pbr`

Now it is used only from tox in order to create the fancy auto-generated "ChangeLog" based on commits and tags

So, it should not be a problem

Could you check if it fix your current issue

FYI we will copy many things from that project to pylint-odoo

I have checked pylint, flake8, pre-commit and other projects and they are using the out-of-the-box packages way

Issue #331, opened 2021-05-18
Pylint crashes when object is str


Yenthe666:
Steps to reproduce:
1. Clone `git clone https://github.com/OCA/rma`
2. Checkout `10.0`
3. Run Pylint from the terminal on the module `crm_claim_ext`:
```
pylint --load-plugins=pylint_odoo --disable=C -f json crm_claim_ext
```

Result:
```
Traceback (most recent call last):
  File "/usr/local/bin/pylint", line 11, in <module>
    sys.exit(run_pylint())
  File "/usr/local/lib/python3.7/site-packages/pylint/__init__.py", line 22, in run_pylint
    PylintRun(sys.argv[1:])
  File "/usr/local/lib/python3.7/site-packages/pylint/lint/run.py", line 344, in __init__
    linter.check(args)
  File "/usr/local/lib/python3.7/site-packages/pylint/lint/pylinter.py", line 871, in check
    self.get_ast, self._iterate_file_descrs(files_or_modules)
  File "/usr/local/lib/python3.7/site-packages/pylint/lint/pylinter.py", line 904, in _check_files
    self._check_file(get_ast, check_astroid_module, name, filepath, modname)
  File "/usr/local/lib/python3.7/site-packages/pylint/lint/pylinter.py", line 930, in _check_file
    check_astroid_module(ast_node)
  File "/usr/local/lib/python3.7/site-packages/pylint/lint/pylinter.py", line 1063, in check_astroid_module
    ast_node, walker, rawcheckers, tokencheckers
  File "/usr/local/lib/python3.7/site-packages/pylint/lint/pylinter.py", line 1107, in _check_astroid_module
    walker.walk(ast_node)
  File "/usr/local/lib/python3.7/site-packages/pylint/utils/ast_walker.py", line 72, in walk
    callback(astroid)
  File "/usr/local/lib/python3.7/site-packages/pylint_odoo/misc.py", line 217, in visit_module
    self.wrapper_visit_module(node)
  File "/usr/local/lib/python3.7/site-packages/pylint_odoo/misc.py", line 203, in wrapper_visit_module
    if not check_method():
  File "/usr/local/lib/python3.7/site-packages/pylint_odoo/checkers/modules_odoo.py", line 871, in _check_file_not_used
    set(self._get_xml_referenced_files())
  File "/usr/local/lib/python3.7/site-packages/pylint_odoo/checkers/modules_odoo.py", line 843, in _get_xml_referenced_files
[...]

PR #343, opened 2021-09-08
[FIX] xml-syntax-error: Fix crash if a xml has syntax errors

moylop260:
Fix https://github.com/OCA/pylint-odoo/issues/331

### DIFF
--- a/pylint_odoo/checkers/modules_odoo.py
+++ b/pylint_odoo/checkers/modules_odoo.py
@@ -7,8 +7,8 @@ import re
 import astroid
 import polib
 from collections import defaultdict
+from lxml import etree
 from pylint.checkers import utils
-from six import string_types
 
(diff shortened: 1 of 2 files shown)


moylop260:
After this fix running the following command:
 - `pylint -d all -e odoolint --load-plugins=pylint_odoo /tmp/rma/crm_claim_ext/ |grep crm_claim_ext_view.xml`

For https://github.com/OCA/rma/blob/d8d1bdbd9c09c74f9cd1ad12e13db884712cc9f3/crm_claim_ext module

The result is:
 - rma/crm_claim_ext/crm_claim_ext_view.xml:1:0: E7902:  error: Opening and ending tag mismatch: page line 30 and field, line 33, column 21 (crm_claim_ext_view.xml, line 33) (xml-syntax-error)

instead of crash

Issue #323, opened 2021-03-26
modify dangerous-view-replace-wo-priority: Replace in QWeb template is not detected


luisg123v:
The lint is not detecting replaces when they're inside QWeb templates 
`<template>`
instead of
`<record ... model="ir.ui.view">`
for instance:

```
<template id="some_qweb_template" inherit_id="web.some_qweb_template">
    <xpath expr="//div[@name='some_div']" position="replace">
        ...
    </xpath>
</template>
```

luisg123v:
CC @moylop260 @carmenmiranda

PR #347, opened 2021-09-09
[ADD] dangerous-qweb-replace-wo-priority: Consider view templates

moylop260:
Fix https://github.com/OCA/pylint-odoo/issues/323

### DIFF
--- a/pylint_odoo/checkers/modules_odoo.py
+++ b/pylint_odoo/checkers/modules_odoo.py
@@ -105,6 +105,12 @@ ODOO_MSGS = {
         'dangerous-view-replace-wo-priority',
         settings.DESC_DFLT
     ),
+    'W%d41' % settings.BASE_OMODULE_ID: (
+        '%s Dangerous use of "replace" from view '
+        'with priority %s < %s',
+        'dangerous-qweb-replace-wo-priority',
+        settings.DESC_DFLT
+    ),
     'W%d30' % settings.BASE_OMODULE_ID: (
         '%s not used from manifest',
         'file-not-used',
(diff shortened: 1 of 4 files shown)


moylop260:
@OCA/community-maintainers @OCA/board @OCA/oca-contributors 

We had a false positive for views of type `<template...`

Releasing this change could put red a few projects if there is `replace` without `priority`

More info about this check:
 - https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#inherited-xml

Could you review it, please?

pylint_odoo/checkers/modules_odoo.py
```diff
@@ -853,16 +853,31 @@ def _check_dangerous_view_replace_wo_priority(self):
...
                 priority = self._get_priority(view)
                 is_replaced_field = self._is_replaced_field(view)
                 if is_replaced_field and priority < self.config.min_priority:
                     self.msg_args.append((
                         "%s:%s" % (xml_file, view.sourceline), priority,
                         self.config.min_priority))
+
+            # view template
+            xml_doc = self.parse_xml(xml_file_path)
+            for template in xml_doc.xpath("/odoo//template|/openerp//template"):
```
hbrunn, line 869: how about the `data` node?

moylop260: The double `//` considers `data` node

The test_repo has the `data` node examples and it is detected

Check the following example:

 - https://github.com/OCA/pylint-odoo/blob/1ee7a910bdd51627c067a88da44178853cd1ca9c/pylint_odoo/test_repo/broken_module/model_view_odoo2.xml#L3


And check the following output:
 - https://app.travis-ci.com/github/OCA/pylint-odoo/jobs/536551020#L408

`pylint_odoo/test_repo/broken_module/model_view_odoo2.xml:17: [W7902(duplicate-xml-record-id), ]  Duplicate xml record id "data/view_model_form2_noupdate_0" in model_view2.xml:5`

It based on the following similar xpath:
 - https://github.com/OCA/pylint-odoo/blob/1ee7a910bdd51627c067a88da44178853cd1ca9c/pylint_odoo/misc.py#L536

Check the following documentation about xpath:
  - https://dev.to/masihurmaruf/locator-strategy-xpath-54p7

```txt
A single slash ‘/’ anywhere in XPath signifies to look for the element immediately inside its parent element.
A double slash ‘//’ signifies to look for any child or any grand-child (descendant) element inside the parent element.
```

moylop260:
I just added a new check in order to enable it explicitly (or disable it) and avoid raising errors for old green repositories

pylint_odoo/checkers/modules_odoo.py
dreispt: I believe the root can be a `<data>` element instead of `<odoo>`. Maybe this is what Holger was referring to.

moylop260: I think root can't be <data> since that not all xml have <data>

Check the following example
https://github.com/odoo/odoo/blob/14.0/addons/account/views/account.xml

Notice the root node is <odoo>

@hbrunn meant that we should consider odoo and data but it is already considered

Is not?

hbrunn: yes that's what I meant and it's indeed covered by the double slash I didn't consider

Issue #357, opened 2021-11-27
[BUG] missing-readme triggers on repos with only one module that has migrations scripts


gaikaz:
## Describe the bug
pylint-odoo "thinks" migration folders are Odoo modules when there is only one module in a repo that has a migration script.
And then it gives a `missing-readme` fail on that migration folder.

## To Reproduce
Best quick/easy example I could come up is:
1. clone https://github.com/OCA/maintenance
2. Checkout `15.0` branch
3. Locally delete all modules except for `maintenance_equipment_sequence` (it has migration script)
4. Run `pre-commit run -a`

Or just try to reproduce it having a single module with a migration script.

**Expected behavior**
Pre-commit checks result being all green

moylop260:
@gaikaz 

I couldn't reproduced it locally

In fact, the pre-commit CI is green:
 - https://github.com/OCA/maintenance/runs/4233642003?check_suite_focus=true#step:4:75

Steps I did locally:
```bash
git clone git@github.com:OCA/maintenance.git -b 15.0 /tmp/maintenance && \
     cd /tmp/maintenance && \
     pre-commit run --all
```

Output:

```txt
flake8...................................................................Passed
pylint with optional checks..............................................Passed
- hook id: pylint
- duration: 1.41s
pylint with mandatory checks.............................................Passed
```

Could you double confirm the steps to reproduce it, please?

gaikaz:
@moylop260 
Thank you for actually giving this a go.

About the steps, yes, you seem to have skipped step no. 3: 
> ... delete all modules except for `maintenance_equipment_sequence` ...

So the one-liner should be:
```bash
git clone git@github.com:OCA/maintenance.git -b 15.0 /tmp/maintenance &&\
    cd /tmp/maintenance && \
    rm -r */ &&\
    git checkout maintenance_equipment_sequence &&\
    pre-commit run --all
```

This will give generate the `setup` folder again - that's expected. You could try running the pre-commit twice to make it cleaner. But the issue then appears as an output that looks like this:
```
flake8...................................................................Passed
pylint with optional checks..............................................Passed
- hook id: pylint
- duration: 0.6s

************* Module pre-migration
maintenance_equipment_sequence/migrations/15.0.1.0.0/pre-migration.py:1: [C7902(missing-readme), ] Missing ./README.rst file. Template here: https://github.com/OCA/maintainer-tools/blob/master/template/module/README.rst

pylint with mandatory checks.............................................Failed
- hook id: pylint
[...]

PR #358, opened 2021-12-01
[FIX] OdooChecker: Discard migrations folder as odoo main module

moylop260:
pre-commit script is running file by file as:

 - `pylint module/__init__.py module/migrations/pre-commit.py`

It is generating a corner-case where odoo manifest file variable is filled

and pre-commit is the main module and it was detecting as main odoo module

But it is not

Validating the name of node.file is `__init__.py`
we can be sure that it is related with the odoo main module

Fix https://github.com/OCA/pylint-odoo/issues/357

### DIFF
--- a/pylint_odoo/misc.py
+++ b/pylint_odoo/misc.py
@@ -230,7 +230,7 @@ class PylintOdooChecker(BaseChecker):
             self.manifest_dict = {}
             self.manifest_file = None
         self.is_main_odoo_module = False
-        if self.manifest_file and (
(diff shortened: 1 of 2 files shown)


pedrobaeza (approved):

gaikaz (approved):

gaikaz:
@moylop260 
Could we please version bump, so this fix can be used in pre-commit config?
Or am I missing some obvious way to do it already?

moylop260:
Done!

PR #362, opened 2022-02-03
[FIX] Incorrect missing-return for tearDownClass

rven:


moylop260:
Thank you for your contribution!

jwaes:
I don't get it ... 
What is wrong with this code and what should it return ?

```
    @classmethod
    def tearDownClass(cls):
        cls.loader.restore_registry()
        super(TestBaseRevision, cls).tearDownClass()
```

gaikaz:
@jwaes Probably means like this, but I'm not sure if in that context this makes sense.
```python
    @classmethod
    def tearDownClass(cls):
        cls.loader.restore_registry()
        return super(TestBaseRevision, cls).tearDownClass()
```

pedrobaeza:
Yes, always return the possible upstream value. It doesn't matter if it's none. Do that methodically.

rven:
> I don't get it ... What is wrong with this code and what should it return ?
> 
> ```
>     @classmethod
>     def tearDownClass(cls):
>         cls.loader.restore_registry()
>         super(TestBaseRevision, cls).tearDownClass()
> ```

It shouldn't return anything, that's the fix I made here.

pedrobaeza:
Well, I don't see a problem to return `None`.

jwaes:
Anyway ... without the return, the pre-commit was still giving issues.

https://github.com/jwaes/server-ux/commit/352663ac52a0a44408f76050ca16ea2881ec7551

pedrobaeza:
Probably that's because the pylint-odoo version (pinned) used there is not including this fix. But as said, I don't see this patch as needed. It's a good habit to always return upstream return value.

PR #364, opened 2022-04-01
[IMP] sql_injection: Detect possible sql injections when using fstrings

### LINKED ISSUE #363: [REF] sql-injection: Support f-string
@xmo-odoo added support for f-string in the test_lint module of Odoo:
 - https://github.com/odoo/odoo/pull/81639

We need to support this feature too for pylint-odoo

Consider add unittest

antonag32:
Supposed to fix #363. This is a work in progress as there are still some things to complete, like:

- [x] Detect use of fstrings inside `execute` calls
- [x] Do not flag fstrings which only use constants as possible sql injections
- [x] Write unit tests 

And perhaps some other things which will come up during development.

### DIFF
--- a/pylint_odoo/test/main.py
+++ b/pylint_odoo/test/main.py
@@ -1,8 +1,7 @@
-
 import os
 import stat
 import sys
-from tempfile import gettempdir
+from tempfile import gettempdir, NamedTemporaryFile
 import six
 
 import unittest
@@ -431,6 +430,42 @@ class MainTest(unittest.TestCase):
         real_errors = pylint_res.linter.stats['by_msg']
         expected_errors = {}
         self.assertDictEqual(real_errors, expected_errors)
+
+    @unittest.skipUnless(
+        sys.version_info >= (3, 6),
+        "Fstrings (PEP498) are only supported since 3.6"
+    )
+    def test_145_check_fstring_sqli(self):
+        """Verify the linter is capable of finding SQL Injection vulnerabilities
+        when using fstrings.
+        Related to https://github.com/OCA/pylint-odoo/issues/363"""
+        extra_params = [
+            '--disable=all',
+            '--enable=sql-injection'
+        ]
+        queries = '''
+def fstring_sqli(self):
+   self.env.cr.execute(f"SELECT * FROM TABLE WHERE SQLI = {self.table}")
+   self.env.cr.execute(
+       f"SELECT * FROM TABLE WHERE SQLI = {'hello' + self.name}"
+   )
+   self.env.cr.execute(f"SELECT * FROM {self.name} WHERE SQLI = {'hello'}")
+   death_wish = f"SELECT * FROM TABLE WHERE SQLI = {self.name}"
+   self.env.cr.execute(death_wish)
+def fstring_no_sqli(self):
+   self.env.cr.execute(f"SELECT * FROM TABLE WHERE SQLI = {'hello'}")
+   self.env.cr.execute(
+       f"CREATE VIEW {self._table} AS (SELECT * FROM res_partner)"
+   )
+   self.env.cr.execute(f"SELECT NAME FROM res_partner LIMIT 10")
+           '''
+        with NamedTemporaryFile(mode='w') as f:
+            f.write(queries)
+            f.flush()
+            pylint_res = self.run_pylint([f.name], extra_params)
+
+        real_errors = pylint_res.linter.stats['by_msg']
+        self.assertDictEqual(real_errors, {'sql-injection': 4})
 
 
 if __name__ == '__main__':
--- a/pylint_odoo/checkers/no_modules.py
+++ b/pylint_odoo/checkers/no_modules.py
@@ -441,6 +441,11 @@ class NoModuleChecker(misc.PylintOdooChecker):
         # sql.SQL or sql.Identifier is OK
         if self._is_psycopg2_sql(node):
             return True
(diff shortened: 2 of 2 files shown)


antonag32:
@moylop260

antonag32:
@moylop260 
The idea to allow fstrings that only use constants came from your commit 5d9878a35512 which allowed calls like:
```
self.env.cr.execute("SELECT * FROM %s" % 'table_constant')
```
But after actually checking it, you can't even do:
```
self.env.cr.execute(f'SELECT * FROM {'table'} WHERE name = {self.name}')
```

So perhaps it makes sense to just flag all fstrings as evil and throw an sql error.

antonag32:
Implemented a recursive strategy to check for constants inside fstrings and mark them as safe if all formatted values are made up of constants only. Probably no one will use it in real life but its the technically correct thing to do. 

Added a unit test to verify SQLI detection is working as corrected.

antonag32:
As of today, 4 tests are failing on the master branch when I run the whole suit.

- test_20_expected_errors
- test_30_disabling_errors
- test_70_without_jslint_installed
- test_80_with_jslint_error

They still fail after my changes, but no other tests failed, so I assume my work didn't break previous work.

moylop260:
I think you will need to bump the sql-injection in the `EXPECTED_ERRORS` dict:
 - https://github.com/OCA/pylint-odoo/blob/80442ff58ff22810929f949fe496fce7223abf75/pylint_odoo/test/main.py#L61

moylop260:
And fix the lints, please

```txt
$ flake8 --ignore=E722,F601,F841,W503,W504,W605 --max-line-length=88 --exclude=__init__.py .
./pylint_odoo/test_repo/broken_module/models/broken_model.py:501:68: E999 SyntaxError: invalid syntax
./pylint_odoo/test_repo/broken_module/models/broken_model.py:539:89: E501 line too long (105 > 88 characters)
./pylint_odoo/test/main.py:436:89: E501 line too long (107 > 88 characters)
```

antonag32:
@moylop260. Done, I bumped the sql-injection in `EXPECTED_ERRORS` and fixed the linting warnings, although locally I was not able to get the E999 SyntaxError warning.

antonag32:
@moylop260 Since fstrings are not supported until Python 3.6 I made a separate test and embedded data int it instead of modifying test repos. All fstring testing is now done inside a single test which will be skipped if the version is not at least 3.6.

antonag32:
@moylop260 
Apparently I nested existing code inside my checks, which meant certain injections were not being detected. It is fixed now. `test_20_expected_errors` still blows up but sql-injections are no longer one of the reasons.

pylint_odoo/test/main.py
```diff
@@ -432,6 +433,42 @@ def test_140_check_migrations_is_not_odoo_module(self):
...
+            def fstring_no_sqli(self):
+                self.env.cr.execute(f"SELECT * FROM TABLE WHERE SQLI = {'hello'}")
+                self.env.cr.execute(
+                    f"CREATE VIEW {self._table} AS (SELECT * FROM res_partner)"
+                )
+                self.env.cr.execute(f"SELECT NAME FROM res_partner LIMIT 10")
+            '''
+        with NamedTemporaryFile(mode='w', encoding='utf-8', delete=False) as f:
+            self.addCleanup(os.remove, f.name)
+            f.write(dedent(queries).lstrip())
```
moylop260, line 466: What about?

```suggestion
        with NamedTemporaryFile(mode='w', encoding='utf-8') as f:
            f.write(queries)
            f.flush()
            pylint_res = self.run_pylint([f.name], extra_params)
```

You even can use the string as you need it without calling extra string formatters:
```python
        queries = '''
def fstring_sqli(self):
    self.env.cr.execute(f"SELECT * FROM TABLE WHERE SQLI = {self.table}")
    self.env.cr.execute(
        f"SELECT * FROM TABLE WHERE SQLI = {'hello' + self.name}"
    )
    self.env.cr.execute(f"SELECT * FROM {self.name} WHERE SQLI = {'hello'}")
    death_wish = f"SELECT * FROM TABLE WHERE SQLI = {self.name}"
    self.env.cr.execute(death_wish)
def fstring_no_sqli(self):
    self.env.cr.execute(f"SELECT * FROM TABLE WHERE SQLI = {'hello'}")
    self.env.cr.execute(
        f"CREATE VIEW {self._table} AS (SELECT * FROM res_partner)"
    )
    self.env.cr.execute(f"SELECT NAME FROM res_partner LIMIT 10")'''
```

moylop260 (changes requested):

pylint_odoo/test/main.py
```diff
@@ -432,6 +433,42 @@ def test_140_check_migrations_is_not_odoo_module(self):
         expected_errors = {}
         self.assertDictEqual(real_errors, expected_errors)
 
+    @unittest.skipUnless(
+        sys.version_info[0:2] >= (3, 6),
```
moylop260, line 437: I think you even don't need index range here [0:2]

```suggestion
        sys.version_info >= (3, 6),
```

antonag32: True, I'll remove it.

pylint_odoo/test/main.py
antonag32: I used dedent so the code looked prettier (IMO). I'll make the changes.

antonag32:
@moylop260 All requested changes were attended.

PR #365, opened 2022-04-06
[FIX] Pinned lazy-object-proxy for Python 2.7

antonag32:
Travis builds on Python 2.7 break because `lazy-object-proxy` can't be installed. This should fix it.


moylop260:
please, could you fix the another CI error?

https://app.travis-ci.com/github/OCA/pylint-odoo/jobs/566194574#L272

PR #310, opened 2021-01-01
[ADD] test-folder-imported: check if tests folder is imported in init file

fernandahf:
Description of the issue/feature this PR addresses:

 Currently, the `__init__.py` file in the root directory of an Odoo module may import the tests folder, e.g. https://github.com/OCA/server-tools/blob/cca5dffd0dec68e1aa4209a8747d4381a97efa89/ir_sequence_standard_default/__init__.py#L6

Current behavior before PR: 

Folder test imported in Odoo module does not check.

Desired behavior after PR is merged:

Folder test imported in Odoo module is check.

Fix #293 

-- I confirm I have signed the CLA and read the PR guidelines at www.odoo.com/submit-pr

PR #373, opened 2022-05-25
[IMP] test-folder-imported: Consider 'from . import test' missing case

moylop260:
The original PR:
 - https://github.com/OCA/pylint-odoo/pull/310

Talks about the following case:
 - https://github.com/OCA/server-tools/blob/cca5dffd0dec68e1aa4209a8747d4381a97efa89/ir_sequence_standard_default/__init__.py#L6

It is using "from . import tests"

But this case of use was not used in the unittest

So, it is not detecting this common case


luisg123v (approved):
LGTM but shouldn't it be IMP instead of REF?

PR #374, opened 2022-05-25
[IMP] duplicate-po-message-definition: Detecting duplicated only with msgid

moylop260:
Odoo raises error if the msgid is duplicated even if there are not models entries not duplicated

So, it is considering these kind of cases in order to match with Odoo errors but raising from CI early with static checker

Original PR:
 - https://github.com/OCA/pylint-odoo/pull/290

### DIFF
--- a/pylint_odoo/checkers/modules_odoo.py
+++ b/pylint_odoo/checkers/modules_odoo.py
@@ -556,10 +556,7 @@ class ModuleChecker(misc.WrapperModuleChecker):
             for entry in po:
                 if entry.obsolete:
                     continue
-                # Using `set` in order to fix false red
-                # if the same entry has duplicated occurrences
-                for occurrence in set(entry.occurrences):
-                    duplicated[(hash(entry.msgid), hash(occurrence))].append(entry)
+                duplicated[hash(entry.msgid)].append(entry)
             for entries in duplicated.values():
                 if len(entries) < 2:
                     continue
--- a/pylint_odoo/test/main.py
+++ b/pylint_odoo/test/main.py
@@ -27,7 +27,7 @@ EXPECTED_ERRORS = {
     'deprecated-openerp-xml-node': 5,
     'development-status-allowed': 1,
     'duplicate-id-csv': 2,
-    'duplicate-po-message-definition': 2,
+    'duplicate-po-message-definition': 3,
     'duplicate-xml-fields': 9,
     'duplicate-xml-record-id': 2,
     'external-request-timeout': 47,
--- a/pylint_odoo/test_repo/broken_module/i18n/es.po
+++ b/pylint_odoo/test_repo/broken_module/i18n/es.po
@@ -12,8 +12,13 @@ msgstr ""
 "Plural-Forms: \n"
 
 #. module: broken_module
-#: model:ir.model.fields,field_description:broken_module.field_description
 #: model:ir.model.fields,field_description:broken_module.field_wizard_description
+#, python-format
+msgid "Branch"
+msgstr ""
+
+#. module: broken_module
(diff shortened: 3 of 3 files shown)


luisg123v (approved):
LGTM :+1:

moylop260:
Hi @hbrunn 

I'm not sure since that the PO raises error when it is duplicated and Odoo too

@luisg123v could you double-confirm this, please?

hbrunn:
My commit is https://github.com/moshchot/BANKayma/pull/16/commits/7e0f64d6063688108ad14dac9382fcfec1d29c65#diff-586ca03518299c22fd35e292a1a5d33b5405e9e238298570cac682941938c2e2L561-R571

(so we need to translate 'Draft' differently in some contexts)

which [fails](https://github.com/moshchot/BANKayma/actions/runs/7738062882/job/21098160839) on this test, but the Odoo test works fine: https://github.com/moshchot/BANKayma/actions/runs/7738062884/job/21098161098

moylop260:
@hbrunn 

Sorry for delay

I just tested this case of use

I  have applied the following patch:

```patch
diff --git a/addons/lunch/i18n/es_419.po b/addons/lunch/i18n/es_419.po
index 9939985e..d5893dbd 100644
--- a/addons/lunch/i18n/es_419.po
+++ b/addons/lunch/i18n/es_419.po
@@ -397,13 +397,34 @@ msgstr "Acción requerida"

 #. module: lunch
 #: model:ir.model.fields,field_description:lunch.field_lunch_alert__active
+msgid "Active"
+msgstr "Activo 1"
+
+#. module: lunch
 #: model:ir.model.fields,field_description:lunch.field_lunch_order__active
+msgid "Active"
+msgstr "Activo 2"
+
+#. module: lunch
 #: model:ir.model.fields,field_description:lunch.field_lunch_product__active
+msgid "Active"
+msgstr "Activo 3"
+
+#. module: lunch
 #: model:ir.model.fields,field_description:lunch.field_lunch_product_category__active
+msgid "Active"
+msgstr "Activo 3"
+
+#. module: lunch
 #: model:ir.model.fields,field_description:lunch.field_lunch_supplier__active
+msgid "Active"
+msgstr "Activo 4"
+
+#. module: lunch
 #: model_terms:ir.ui.view,arch_db:lunch.lunch_alert_view_search
 msgid "Active"
-msgstr "Activo"
+msgstr "Activo 5"
+

 #. module: lunch
 #: model:ir.model.fields,field_description:lunch.field_lunch_supplier__activity_ids

```

I start a new database:
`~/instance/odoo/odoo-bin -d lunch -i lunch --stop-after-init --load-language="es_MX"`

I notice the following logger:
`...odoo.addons.base.models.ir_module: module lunch: loading translation file /home/odoo/instance/odoo/addons/lunch/i18n/es_419.po for language es_MX`

Then I just get the fields values translated:

```python
fields = self.env["ir.model.fields"].with_context(lang="es_MX").search([("model", "like", "lunch%"), ("name", "=", "active")])
fields.mapped("field_description")
```

The result is:
`['Activo 5', 'Activo 5', 'Activo 5', 'Activo 5', 'Activo 5']`

It means, that even if Odoo is not raising an error for new versions the translated value is using only the last one for all items

It is not an odoo deficiency but po files

In fact, `gettext` parsing raises the same error as this lint

hbrunn:
thanks for investigating. For the time being I just used different terms, but there are places where this is awkward in English then.

I wonder what the general solution here is. Use id-like texts in code and add an en_US.po file?

yostashiro:
@hbrunn I don't have a solution but here is a related discussion: https://github.com/odoo/odoo/issues/136598

PR #376, opened 2022-06-10
[PERF] Checkers only run when enabled #372

### LINKED ISSUE #372: [PERF] use is_message_enabled for all lints
Currently, there are some lints that are not using the following validation:

https://github.com/OCA/pylint-odoo/blob/5c15bca5fa12f3eaef54e03a1d8718150f6f2d5a/pylint_odoo/checkers/no_modules.py#L872

That generates all code in those lints, even if they are disabled, are checked.

Example: 

https://github.com/OCA/pylint-odoo/blob/5c15bca5fa12f3eaef54e03a1d8718150f6f2d5a/pylint_odoo/checkers/no_modules.py#L593-L751

antonag32:
Fixes #372, this is still a WIP as some files still need to be processed

### DIFF
--- a/pylint_odoo/checkers/modules_odoo.py
+++ b/pylint_odoo/checkers/modules_odoo.py
@@ -181,9 +181,8 @@ ODOO_MSGS = {
     ),
 }
 
-
 DFTL_README_TMPL_URL = 'https://github.com/OCA/maintainer-tools' + \
-    '/blob/master/template/module/README.rst'
+                       '/blob/master/template/module/README.rst'
 DFTL_README_FILES = ['README.rst', 'README.md', 'README.txt']
 DFTL_MIN_PRIORITY = 99
 # Files supported from manifest to convert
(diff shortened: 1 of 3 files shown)


pylint_odoo/checkers/modules_odoo.py
```diff
@@ -364,11 +365,13 @@ def _get_odoo_module_imported(self, node):
                     odoo_module.append(packages[2])
         return odoo_module
 
+    @utils.check_messages('odoo-addons-relative-import')
```
moylop260, line 368: FYI modules_odoo is not using/processing this decorator (yet)

antonag32: Do you mean the decorators have no effect and the method will still run? Which means I should use `self.linter.is_message_enabled`?

moylop260: yes

moylop260: and so?

antonag32:
@moylop260 Removed all decorators from `modules_odoo` since they have no effect currently

moylop260 (approved):
👍

PR #377, opened 2022-06-13
[FIX] translation-positional-used: Limit only for >=14.0

moylop260:
<=13.0 does not have positional arguments:
 - https://github.com/odoo/odoo/blob/a03aa23461498bba67/odoo/tools/translate.py#L447

>=14.0 has positional arguments:
 - https://github.com/odoo/odoo/blob/6a6e5f92789e1c/odoo/tools/translate.py#L446

So, we only are limiting this check for odoo >=14.0 since past version is not already valid


moylop260:
This check is not related with using parameters in `_` method

I will create a new issue to prefer `translation-not-lazy` similar to logging-not-lazy

PR #379, opened 2022-06-16
[FIX] missing-import-error: add xlrd to whitelist

luisg123v:
The `xlrd` package is in Odoo's requirements since v10 [1].

[1] https://github.com/odoo/odoo/commit/0ac303d12ff0


luisg123v:
Hi @moylop260,

Could you review, please?

CC @keylor2906 @fernandahf

PR #385, opened 2022-06-27
[REF] tests: Add unittest to check only one check enabled

moylop260:


moylop260:
@antonag32 

Could you check why this new unittest is failing only with https://github.com/OCA/pylint-odoo/commit/01919a0ac6f02e2bc8fc929a096e9d26045bd2ed but not for previous one, please?

PR #388, opened 2022-08-08
[REF] external-request-timeout: Add ftplib.FTP method

moylop260:
It is better using timeout for this method too as documentation says:
 - https://docs.python.org/es/3/library/ftplib.html#ftplib.FTP_TLS


luisg123v:
like -> as
"like" is for similarity/comparisons, while "as" would mean "according to"

PR #392, opened 2022-09-30
[FIX] Only adds W8113 when field string is the same as title() format

### LINKED ISSUE #387: [BUG][W8113]String parameter equal to name of variable (capital letters)
Related to: https://github.com/OCA/knowledge/pull/357#discussion_r930728422

**Use case**:

There is a field
```
has_changes_pending_approval = fields.Boolean(
        string="Has changes pending approval"
)
```
The message "_The attribute string is redundant. String parameter equal to name of variable_" (W8113) is detected but it is incorrect because the string that the field had did NOT contain capital letters.

**Expected behavior**:
Only show the message if the existing string is exactly the same as the name of the variable with capital letters.

hugosantosred:
This allows to have titles in different case.

For example this field declaration is not going to set a W8113:

```
has_changes_pending_approval = fields.Boolean(
        string="Has changes pending approval"
)
```

But if we define the field in this way it will set a W8113:

```
has_changes_pending_approval = fields.Boolean(
        string="Has Changes Pending Approval"
)
```

Fixes #387

### DIFF
--- a/pylint_odoo/checkers/no_modules.py
+++ b/pylint_odoo/checkers/no_modules.py
@@ -659,8 +659,8 @@ class NoModuleChecker(misc.PylintOdooChecker):
                 if (not is_related and isinstance(argument, astroid.Const) and
                     (index ==
                      FIELDS_METHOD.get(argument.parent.func.attrname, 0)) and
-                    (argument.value in
-                     [field_name.capitalize(), field_name.title()])):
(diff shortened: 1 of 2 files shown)


moylop260:
👍 

(py3.7 fails expected)

pedrobaeza:
Thanks!

PR #422, opened 2022-10-24
DRAFT: [REF] pylint_odoo: Process (max|min) odoo version

moylop260:
- [ ] Look for a check only valid for one particular version and test the feature


moylop260:
Superseded by https://github.com/OCA/pylint-odoo/pull/424

In the future we could use this way in order to know if the check is enabled early

But it will require monkey patches

Issue #356, opened 2021-10-05
New linter: Don't `write` on a computed field


pedrobaeza:
Or bad things will happen

Example:

https://github.com/OCA/stock-logistics-warehouse/blob/9f0124ade73a167c7f512f7f4aef57e6a4ffceb0/sale_stock_available_info_popup/models/sale_order.py#L39

Immediate solution: do `update` instead.

moylop260:
Hi Pedro

Thank you for creating this issue

I have been this lint in my mind a long time ago

FedericoGregori:
Sorry I don't understand the "Or bad things will happen". Can I ask an example? This is a huge thing to teach to every dev I think.

pedrobaeza:
Onchange methods are executed in "edit-mode", so you need to save to get the definitive values saved on the DB, but if you do a `write` on it, that changes are immediately saved to the DB, so if you discard the changes on UI, you will have inconsistent data. That's why you have to do `update` instead, that only modifies the temporary dataset.

The same applies with `create` and with `unlink`.

moylop260:
> It is hard to distinguish `write` on record or non-record object.

Yes, it is hard.

Instead, We could check all the fields with compute parameter and check if that method has a `write`

I mean,

name = fields.Char(compute="_compute_name")

Then get the method "_compute_name" and check if is using ".write("

moylop260:
@antonag32 Could you check it, please?

antonag32:
@moylop260 Yes, will start working on it

voronind:
> In this case the `write` is a method of a built-in object `file`
> 
> Ast detects this so we could discard all detected ones

OK. And in this case?

```python
def _compute_name(self):
    ...
    unknown_type_object = self._get_object()
    unknown_type_object.write(b'some')
```

moylop260:
It should came from a browse variable or only self

moylop260:
I was testing the checker and the following line was detected in Odoo core

Do you know if there is a valid case to use write or could be it an valid error?
 - https://github.com/odoo/odoo/blob/d8d47f9ff8554f4b39487fd2f13c153c7d6f958d/addons/product_margin/models/product_product.py#L179

moylop260:
New findings updated from odoo repos

odoo/odoo@16.0

```txt
addons/fleet/models/fleet_vehicle.py:141:12: E8135: Compute method calling `write`. Use `update` instead. (no-write-in-compute)
[...]

PR #427, opened 2022-10-25
[ADD] no-write-in-compute: Don't write on a computed field 

moylop260:
Fix https://github.com/OCA/pylint-odoo/issues/356

### DIFF
--- a/src/pylint_odoo/checkers/odoo_addons.py
+++ b/src/pylint_odoo/checkers/odoo_addons.py
@@ -1332,7 +1332,7 @@ class OdooAddons(OdooBaseChecker, BaseChecker):
                     self.add_message("no-write-in-compute", node=node_compute_call)
                     continue
                 _root_assignation_node, root_assignation_name = self._get_root_method_assignation(node_compute_call)
-                if root_assignation_name in ["self", "self.env"]:
+                if root_assignation_name in ["self", "self.env", "self.filtered", "self.search", "self.browse"]:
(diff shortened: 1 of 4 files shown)


src/pylint_odoo/checkers/odoo_addons.py
```diff
@@ -170,6 +170,7 @@
         CHECK_DESCRIPTION,
     ),
     "E8130": ("Test folder imported in module %s", "test-folder-imported", CHECK_DESCRIPTION),
+    "E8135": ("Compute method calling 'write' method", "compute-write", CHECK_DESCRIPTION),
```
moylop260, line 173: @pedrobaeza @luisg123v 

Could you help me to define a better and more intuitive message and check-name here, please?

pedrobaeza: I think this can serve:

`"Compute method calling 'write'. Use 'update' instead."`

moylop260: Good!

Thank you!

And name-of-check?

pedrobaeza: `no-write-in-compute`?

moylop260:
@voronind  

I have added all the test-cases that you mentioned in the issue

Please, could you check it

Disclaimer: The code is still work-in-progress but the POC is enough to demo it is possible

PR #437, opened 2022-11-07
[ADD] no-wizard-in-models: Add new check for avoid adding wizards class in models directory

moylop260:
See the structure of modules where wizard models should be in wizards directory instead of models

See the complete structure
https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#complete-structure | C8113

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,6 +42,7 @@ method-required-super | Missing `super` call in "%s" method. | W8106
 method-search | Name of search method should start with "_search_" | C8109
 missing-readme | Missing ./README.rst file. Template here: %s | C8112
 missing-return | Missing `return` (`super` is used) in method %s. | W8110
+no-wizard-in-models | No wizard class for model directory. See the complete structure https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#complete-structure | C8113
 no-write-in-compute | Compute method calling `write`. Use `update` instead. | E8135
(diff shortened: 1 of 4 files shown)


moylop260:
> Guidelines doesn't require to keep Transient models in wizards directory. 

Yes, it is:
 - https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#complete-structure 
   - Screenshot 2022-11-29 at 11 55 53


> Why there is exception for `res.config.settings`? It's wizard model.

Historical reasons

Clone all the OCA repositories and run the following script:
 - `rgrep "res\.config\.settings" ./*/*/models --include="*.py" |grep "models" | wc -l`

You will find ~150 cases

Same for odoo original modules:
 - `rgrep "res\.config\.settings" ./odoo/addons/*/models enterprise/*/models --include="*.py" |grep "models" | wc -l`

You will find ~150 cases too

But it is not common for other cases in OCA repositories
 - `rgrep "class.*TransientModel" ./*/*/models --include="*.py" | grep -vi "settings" |wc -l`
~65 cases

or even for odoo original modules
 - `rgrep "class.*TransientModel" ./odoo/addons/*/models enterprise/*/models --include="*.py" | grep -vi "settings" |wc -l`
~25

Notice this lint is not enabled in OCA

We could say it is a beta check that could be improved or changed even disabled from pylint.cfg

Issue #440, opened 2022-11-07
Failure when installing from source


yajo:
## Describe the bug
Only wheel installation works.

## To Reproduce

```bash
pip install --no-binary :all: pylint-odoo
DEPRECATION: --no-binary currently disables reading from the cache of locally built wheels. In the future --no-binary will not influence the wheel cache. pip 23.1 will enforce this behaviour change. A possible replacement is to use the --no-cache-dir option. You can use the flag --use-feature=no-binary-enable-wheel-cache to test the upcoming behaviour. Discussion can be found at https://github.com/pypa/pip/issues/11453
Collecting pylint-odoo
  Downloading pylint-odoo-8.0.14.tar.gz (39 kB)
  Installing build dependencies ... done
  Getting requirements to build wheel ... error
  error: subprocess-exited-with-error
  
  × Getting requirements to build wheel did not run successfully.
  │ exit code: 1
  ╰─> [23 lines of output]
      Traceback (most recent call last):
        File "/usr/local/lib/python3.11/site-packages/pip/_vendor/pep517/in_process/_in_process.py", line 351, in <module>
          main()
        File "/usr/local/lib/python3.11/site-packages/pip/_vendor/pep517/in_process/_in_process.py", line 333, in main
          json_out['return_val'] = hook(**hook_input['kwargs'])
                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        File "/usr/local/lib/python3.11/site-packages/pip/_vendor/pep517/in_process/_in_process.py", line 118, in get_requires_for_build_wheel
          return hook(config_settings)
                 ^^^^^^^^^^^^^^^^^^^^^
        File "/tmp/pip-build-env-l0vzdgus/overlay/lib/python3.11/site-packages/setuptools/build_meta.py", line 338, in get_requires_for_build_wheel
          return self._get_build_requires(config_settings, requirements=['wheel'])
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        File "/tmp/pip-build-env-l0vzdgus/overlay/lib/python3.11/site-packages/setuptools/build_meta.py", line 320, in _get_build_requires
          self.run_setup()
        File "/tmp/pip-build-env-l0vzdgus/overlay/lib/python3.11/site-packages/setuptools/build_meta.py", line 484, in run_setup
          self).run_setup(setup_script=setup_script)
[...]

PR #445, opened 2022-11-18
[REF] Improved repo structure to facilitate testing

antonag32:
The existence of src was not necessary, and made it difficult to run unittests directly, since the imports depended on [pylint_odoo being installed in the system](https://github.com/OCA/pylint-odoo/blob/1313e2fd159196aaed70396bcf9feb1a1bdac418/tests/test_main.py#L16), instead of importing directly from source code, so tox was the only way to run tests (AFAIK).

After this change, unittests can be directly run without tox, this makes the development process faster, and integrates very well with IDEs that let you run and debug unit tests directly, plus it removes something that is not needed.

pylint itself uses this structure, so I think that is another reason in favor of this change.


antonag32:
This is just a side PR, as I am working on detecting unused python files (not imported but should have), however I wanted to run unittests directly and was not able to which lead me to this. I am aware the REF might be too much for now, but might as well put it out here before I forget about it.

moylop260:
@antonag32 

FYI we are using the following convention related to `src` folder:
 - https://packaging.python.org/en/latest/tutorials/packaging-projects/

Notice `tests` is brother than `src` but I preferred remove them in order to avoid adding to the pypi package heavy test_repo folder then tests one

It is an old and long discussion related to this folder but we have decided to have it TL: DR;
 - https://www.reddit.com/r/learnpython/comments/wpkr4q/comment/ikhdca3/?utm_source=share&utm_medium=web2x&context=3

Also, recently the unittest was passing because it was assuming you are able to import packages that the package released is not

So, it is better an isolated environment (tox) with minimal requirements in order to detect packaging errors early

FYI I have added to README the commands to run the unittest:
 - https://github.com/OCA/pylint-odoo#development

I have detected packaging errors related to this:
 - https://github.com/OCA/odoo-pre-commit-hooks/pull/53#pullrequestreview-1169625666
   - Screenshot 2022-11-18 at 15 12 05

It was fixed with the following code:
 - https://github.com/OCA/odoo-pre-commit-hooks/pull/53/commits/9efddf8e7e403633579327444bdb7d651037d1e5

Similar case for `"ImportError: No module named 'packaging'"`:
 - https://github.com/OCA/pylint-odoo/pull/431

It was a false green because the package `packaging` was installed in the OS environment but using an isolated environment it is detected early

requirements.txt file missing

The last 2 ones was fixed with the following fix:
 - https://github.com/OCA/pylint-odoo/issues/440

So, I even think the src path is good and running the unittest in isolated environment is even better

moylop260:
Also, notice the `update-readme` tox env requires a particular version of python in order to avoid mutable results based on different versions

`--help` output is slightly different for each version of python the isolated environment needs to pin the python version too

antonag32:
Fair enough. Thanks for the context and explanation. I wanted to use the integration Pycharm has by default which lets you run single tests (by just clicking on the green arrow):

Screenshot from 2022-11-18 15-43-05

However they use unittest and the import shown in the original comment fails, no big deal, I will continue using the pytest `-k` flag with Tox.

PR #449, opened 2022-11-23
[ADD] New unused-module check

antonag32:
This new check finds all files that have a class which inherits from one of Odoo's models classes and ensures they are imported.

This is required since classes that interact with the ORM (create models, or inherit them) MUST be imported in the corresponding __init__.py file, otherwise they won't have an effect.


antonag32:
Fixes part of https://github.com/Vauxoo/pylint-odoo/issues/177. Empty files should be checked with a pre-commit hook instead.

antonag32:
@moylop260 Could you review please

antonag32:
Should this be revived? @moylop260 @luisg123v

moylop260:
@luisg123v 

Please check

testing/resources/test_repo/imports_module/__manifest__.py
```diff
@@ -0,0 +1,10 @@
+{
+    'name': 'Test Import Messages',
```
luisg123v, line 2: Are you sure you need a new module?

CAn't you reuse some of the already existing broken modules?

src/pylint_odoo/checkers/__init__.py
```diff
@@ -1,3 +1,4 @@
 from . import odoo_addons
 from . import vim_comment
 from . import custom_logging
+from . import import_checker
```
luisg123v, line 4: Why a new file? Shouldn't you implement new lints into the existing file for Odoo addons?

@moylop260 I don't have enough context to know if this is correct.

luisg123v (changes requested):
Please add comments to the code, code is not trivial and there's not a single comment

src/pylint_odoo/checkers/__init__.py
antonag32: I think odoo_addons is getting too big and grouping lints inside different checkers could help organization.

src/pylint_odoo/checkers/import_checker.py
```diff
@@ -0,0 +1,78 @@
...
+                results.update([name[0] for name in import_stmt.names])
+
+        return results
+
+    @only_required_for_messages("missing-python-import")
+    def visit_classdef(self, node) -> None:
+        odoo_imports = self._get_odoo_import_orm_names(node.parent)
+        if "models" in odoo_imports:
+            not_orm = not any(
+                (basename.startswith("models.") or basename in odoo_imports) for basename in node.basenames
```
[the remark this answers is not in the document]
antonag32: Sorry, what would be the generic check exactly? AFAIK outside of Odoo it is normal to have empty __init_.py files, so I don't think it would be helpful.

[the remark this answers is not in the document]
antonag32: Also I did the models thing to extend it and make sure that both controllers and models are imported in their __init__.py so they make an effect. 

If I add a utility class inside the models folder, it is not required to import it in the __init__.py, I think that is why I checked for class defs.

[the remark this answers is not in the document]
antonag32: I think I understand now, do you mean checking that all .py files are imported somewhere?

testing/resources/test_repo/imports_module/__manifest__.py
antonag32: I think it is cleaner to have test files for each message, or at least have them groupd. pylint does some variation of this and I find it easier to work with, since you don't have to look around different broken modules to find the offending code.

src/pylint_odoo/checkers/import_checker.py
```diff
@@ -0,0 +1,78 @@
...
+    def __init__(self, linter: PyLinter):
+        self.imports = {}
+        self.orm_files = {}
+
+        super().__init__(linter)
+
+    @lru_cache()
+    def _get_odoo_import_orm_names(self, module: Module) -> Set[str]:
+        results = set()
+        for import_stmt in filter(lambda node: isinstance(node, ImportFrom), module.body):
```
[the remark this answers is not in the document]
antonag32: True, I think I  can skip this and just use `visit_importfrom`, will look into it

testing/resources/test_repo/imports_module/__manifest__.py
luisg123v: I think differently, creating a new module will trigger errors in other checkers. Actually, you had to modify unrelated expected errors for this to work.

antonag32:
@moylop260 I have refactored the code to make it more generic and applicable to all files (except those that we agreed to exclude). I also opened an issue on pylint so lets see where that goes.

The pipeline will be all red because I had some trouble running tox, I plan to fix that and tidy up the MR tomorrow.

antonag32:
I had to made a small refactoring to prevent this and future messages from destroying the whole test suite (the new check was triggering on existing broken modules and the new source code was triggering other checkers). As a bonus point messages that don't need a whole module can now use single files (like pylint does).

As for the checker, it is mostly complete, just need to take into account `import` statements as well to have every possible case covered.

antonag32:
@moylop260 @luisg123v 

I think everything is ready. 

1. Works generally (not only for ORM models)
2. Added comments to document inner workings
3. Support for both `from x import y` and `import x` statements
4. Test module to verify all possible cases:
    - tests are excluded
    - migrations as well
    - works for all modules (controllers, models, and everything else)
    - takes into account relative import paths
5. I tested it locally on a production repo from a client and no false positives were raised.

moylop260:
@luisg123v 

Could you review it, please?

src/pylint_odoo/checkers/__init__.py
```diff
@@ -1,3 +1 @@
-from . import odoo_addons
-from . import vim_comment
-from . import custom_logging
+from . import custom_logging, import_checker, odoo_addons, vim_comment
```
luisg123v, line 1: Why now in one line? I think this is not expected

src/pylint_odoo/checkers/import_checker.py
```diff
@@ -0,0 +1,128 @@
...
+from pylint.checkers.utils import only_required_for_messages
+from pylint.lint import PyLinter
+
+from .odoo_base_checker import OdooBaseChecker
+
+ODOO_MSGS = {
+    # C->convention R->refactor W->warning E->error F->fatal
+    "E8401": (
+        "This python module is not imported and has no effect",
+        "unused-module",
```
luisg123v, line 15: IMHO this lint name is not specific enough. What kind of module? Take into account there are Odoo modules and Python modules

I'd say Odoo models instead, e.g. `odoo-model-not-imported`.

What do you think?

src/pylint_odoo/checkers/import_checker.py
```diff
@@ -0,0 +1,128 @@
...
+
+    def __init__(self, linter: PyLinter):
+        self.imports = set()
+        self.modules = set()
+
+        super().__init__(linter)
+
+    @staticmethod
+    def get_module_from_node(node, max_attempts: int = 15) -> Union[None, Module]:
+        """Get the module a node belongs to.
```
luisg123v, line 35: Fix semantic, this is not clear

Did you mean "Return whether the module "a" belongs to the provided node"?

I'm not sure

src/pylint_odoo/checkers/import_checker.py
```diff
@@ -0,0 +1,128 @@
...
+
+        :param node: Node that belongs to the module to be extracted.
+        :param int max_attempts: Amount of attempts that will be made to obtain the module. Nodes that are
```
luisg123v, line 38: amount -> number

src/pylint_odoo/checkers/import_checker.py
```diff
@@ -0,0 +1,128 @@
...
+        nested deeper than max_attempts won't be found.
```
luisg123v, line 39: indent

src/pylint_odoo/checkers/import_checker.py
```diff
@@ -0,0 +1,128 @@
...
+
+        :returns: Module if found, otherwise None.
+        """
+        attempts = 0
+        while attempts < max_attempts:
```
luisg123v, line 44: This may be simplified a bit:

`for attempt in range(max_attempts)`

That way, you won't need to take care of increasing manually the number of attempts

src/pylint_odoo/checkers/import_checker.py
```diff
@@ -0,0 +1,128 @@
...
+
+        return None
+
+    def store_imported_modules(self, node):
+        """Store all modules that are imported by 'from x import y' and 'import x,y,z' statements.
+
+        Relative paths are taken into account for ImportFrom nodes. For example, the following statements
+        would import the following modules (consider the file which contains these statements is module.models.partner):
+
+        from ..wizard import cap, hello --> module.wizard, module.wizard.cap, module.wizard.hello
```
luisg123v, line 62: Indent, to indicate it's code

src/pylint_odoo/checkers/import_checker.py
```diff
@@ -0,0 +1,128 @@
...
+        from ..utils.legacy import db -->  module.utils.legacy, module.utils.legacy.db
+        from . import sale_order --> module.models.sale_order
+
+        import foo --> module.models.foo
+        """
+        module = ImportChecker.get_module_from_node(node)
+        if not module:
+            return
+        if "tests" in module.name:
```
luisg123v, line 71: Join with the above conditional

src/pylint_odoo/checkers/import_checker.py
```diff
@@ -0,0 +1,128 @@
...
+            return
+
+        if isinstance(node, ImportFrom):
+            level = 0 if not node.level else node.level
```
luisg123v, line 75: prefer positive conditions instead of negative ones

It's more readable:

```
if condition:
    x = 1
else:
    x = 2
```

than:

```
if not condition:
    x = 2
else:
    x = 1
```

Because else would be a double negation

src/pylint_odoo/checkers/import_checker.py
```diff
@@ -0,0 +1,128 @@
...
+        elif isinstance(node, Import):
+            level = 1
+        else:
+            raise ValueError("Node must be either ImportFrom or Import")
```
luisg123v, line 79: I don't think you should raise exceptions inside lints.

src/pylint_odoo/checkers/import_checker.py
```diff
@@ -0,0 +1,128 @@
...
+        current_level = 0
+        for char in reversed(module_name):
+            if current_level >= level:
+                break
+            if char == ".":
+                current_level += 1
+
+            slice_index += 1
+
+        if slice_index != 0:
```
luisg123v, line 95:
```suggestion
        if slice_index:
```

src/pylint_odoo/checkers/import_checker.py
```diff
@@ -0,0 +1,128 @@
...
+            root_module = module_name[:-slice_index]
+        else:
+            root_module = module_name
```
luisg123v, line 98: Use inline conditional

luisg123v (changes requested):

luisg123v:
another excluded directory: doc

src/pylint_odoo/checkers/import_checker.py
antonag32: Initially it was for odoo models only but according to the previous review by Moy I made it generic. By Python module I think I am following the official specification which as far as I understand is any `.py` file.

As of now this basically detects unused python files

antonag32:
@luisg123v I have made the requested changes, the only thing left is my question about the code on docstrings.

src/pylint_odoo/checkers/import_checker.py
luisg123v: I consider indenting code is cleaner. CUrrent approach uses several backticks on every line.

luisg123v (changes requested):
This needs local rebase.

antonag32:
I have rebased it. I have also realized I have become a less destructive person since I am double thinking the frozen tests thing. I still think it makes sense to keep test files separately, since that reduces lots of headaches and means adding new messages does not blow up on old test files.

However I wonder if there is any  cleaner way? Maybe it is better to just keep rolling the old way even if it feels "wrong", it seems like the whole test suite was built around the old way.

Issue #452, opened 2022-12-22
Outdated documentation: CLI param --valid-odoo-versions


aleuffre:
## Describe the bug
Since a few versions (EDIT: since this PR: https://github.com/OCA/pylint-odoo/pull/426) the correct CLI command became `--valid-odoo-versions` instead of `--valid_odoo_versions`. The latter now throws an error:

```
pylint: error: Unrecognized option found: valid_odoo_versions=14.0
```

It's simply a matter of updating the docs, I think.
Credit for identifying the issue to BT-rmartin on the OCA Discord/Matrix room.

BT-rmartin:
Thanks @aleuffre. Also mentioning that in the conif file the ODOOLINT section should have all entries with - instead of _ would be welcome.

aleuffre:
@BT-rmartin https://github.com/OCA/pylint-odoo/pull/454#issuecomment-1364030911

PR #454, opened 2022-12-23
Docs: add example of parameter both via CLI and conf file

aleuffre:
https://github.com/OCA/pylint-odoo/issues/452#issuecomment-1363686863


moylop260:
All the parameters from the cli can be generated to configuration file using the following pylint parameter:

`pylint ... --generate-rcfile`

It is documented from pylint directly

So, we don't need to specify cli parameter by parameter to conf file

PR #455, opened 2023-01-07
[FIX] Exclude missing-return checks for computed methods

### LINKED ISSUE #450: [RFC] Ignore missing-return in compute methods
Would it be possible to ignore compute methods for the `missing-return` directive? They aren't meant to return anything and would avoid little absurds like this where the `return` is just superflous:

```python
    discount = fields.Float(compute="_compute_discount", store=True)

    @api.depends("order_id", "order_id.general_discount")
    def _compute_discount(self):
        if hasattr(super(), "_compute_discount"):
            super()._compute_discount()
        for line in self:
            line.discount = line.order_id.general_discount
        return
```

antonag32:
Fixes #450. Compute methods are not supposed to return any value, even if there is a super() in it, since super() shouldn't return anything either.

The only reliable way to ensure a field is computed is if has been declared in the same python module (file), so this exclusion only applies to a limited set of cases.

### DIFF
--- a/src/pylint_odoo/checkers/odoo_addons.py
+++ b/src/pylint_odoo/checkers/odoo_addons.py
@@ -105,8 +105,9 @@ import string
 from collections import Counter, defaultdict
 
 import validators
-from astroid import nodes
+from astroid import Const, nodes
 from pylint.checkers import BaseChecker, utils
+from pylint.lint import PyLinter
 
 from .. import misc
 from .odoo_base_checker import OdooBaseChecker
@@ -361,7 +362,6 @@ PRINTF_PATTERN = re.compile(
 
 
 class OdooAddons(OdooBaseChecker, BaseChecker):
-
     _from_imports = None
     name = "odoolint"
     msgs = ODOO_MSGS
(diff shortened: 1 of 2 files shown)


antonag32:
@luisg123v can you review? thanks

src/pylint_odoo/checkers/odoo_addons.py
```diff
@@ -733,8 +740,16 @@ def _get_assignation_nodes(self, node):
         "translation-field",
         "translation-positional-used",
         "translation-required",
+        "missing-return",
     )
     def visit_call(self, node):
+        # compute method names must be stored to avoid false positives if super is called inside them
+        if self.linter.is_message_enabled("missing-return"):
+            if hasattr(node, "keywords"):
+                for keyword in node.keywords:
```
luisg123v, line 749:
```suggestion
                for keyword in getattr(node, "keywords", []):
```

That way, you don't need the above conditional, so you may avoid an extra indentation level.

Or in fault, combine the two conditionals above

luisg123v (changes requested):

src/pylint_odoo/checkers/odoo_addons.py
antonag32: That's nice, I made the change and also added a break once the `compute` attribute is found since there is no use in looking for a second one AFAIK.

luisg123v (changes requested):
As I commented on the issue, I really don't consider this fix necessary. Always returning super is a good practice and there are other alternatives to sort the onchanges case.

PR #458, opened 2023-05-24
[IMP] add new lint: no-raise-unlink

antonag32:
Closes https://github.com/OCA/odoo-pre-commit-hooks/issues/73.

Context: https://github.com/odoo/odoo/commit/1c8a95809852baefa851985507a07ac7f6465dc7

Raising errors inside unlink(), especially when inherited can leave
stale data behind, therefore they are prohibited. This lint checks
for that.

### DIFF
--- a/tests/test_main.py
+++ b/tests/test_main.py
@@ -63,6 +63,7 @@ EXPECTED_ERRORS = {
     "translation-unsupported-format": 1,
     "use-vim-comment": 1,
     "website-manifest-key-not-valid-uri": 1,
+    "no-raise-unlink": 2,
 }
 
 
@@ -75,11 +76,11 @@ class MainTest(unittest.TestCase):
             "--msg-template={path}:{line} {msg} - [{symbol}]",
             "--rcfile=%s" % os.devnull,
         ]
-        path_modules = os.path.join(
+        self.root_path_modules = os.path.join(
             os.path.dirname(os.path.dirname(os.path.realpath(__file__))), "testing", "resources", "test_repo"
         )
         # Similar to pre-commit way
-        self.paths_modules = glob(os.path.join(path_modules, "**", "*.py"), recursive=True)
+        self.paths_modules = glob(os.path.join(self.root_path_modules, "**", "*.py"), recursive=True)
         self.odoo_namespace_addons_path = os.path.join(
             os.path.dirname(os.path.dirname(os.path.realpath(__file__))),
             "testing",
@@ -147,6 +148,7 @@ class MainTest(unittest.TestCase):
             "translation-too-few-args",
             "translation-too-many-args",
             "translation-unsupported-format",
+            "no-raise-unlink",
         }
         self.default_extra_params += ["--valid-odoo-versions=13.0"]
         pylint_res = self.run_pylint(self.paths_modules)
@@ -165,6 +167,7 @@ class MainTest(unittest.TestCase):
         expected_errors = self.expected_errors.copy()
         expected_errors.update({"manifest-version-format": 6})
         excluded_msgs = {
+            "no-raise-unlink",
             "translation-contains-variable",
         }
         for excluded_msg in excluded_msgs:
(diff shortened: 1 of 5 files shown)


moylop260:
Is it related to issue?

antonag32:
> Is it related to issue?

Yes, but it needs to be moved from oca-pre-commit-hooks:

https://github.com/OCA/odoo-pre-commit-hooks/issues/73

antonag32:
I looked into taking into account functions calls inside unlink that down the line can generate raises, but after some testing I deemed it not practical to do it so I left it as this, it checks for direct `raise` clauses inside unlink functions for ORM models.

moylop260:
@luisg123v 

Please, review

src/pylint_odoo/checkers/function_checker.py
```diff
@@ -0,0 +1,46 @@
...
+            isinstance(parent, ClassDef)
+            and ("_name" in parent.locals or "_inherit" in parent.locals)
+            and node.name == "unlink"
+        )
+
+    @staticmethod
+    def get_enclosing_function(node, depth=10):
+        count = 0
+        parent = getattr(node, "parent", False)
+        while not isinstance(parent, FunctionDef) and count < depth:
```
luisg123v, line 33: I'd suggest using while's when there's no other alternative. In this case, this could be achieved by using

`range(depth)`

On the other hand, why is it required to recursively check parents?

testing/resources/test_repo/eleven_module/res_partner_unlink.py
```diff
@@ -0,0 +1,11 @@
+from odoo import models
+
+
+class ResPartner(models.Model):
+    _inherit = 'res.partner'
+
+    def unlink(self):
+        if self.name == 'explode':
+            raise RuntimeError
```
luisg123v, line 9:
```suggestion
            raise RuntimeError()
```

Same in other exceptions

tests/test_main.py
```diff
@@ -363,6 +364,13 @@ def test_160_check_only_disabled_one_check(self):
             expected_errors.pop(disable_error)
             self.assertDictEqual(real_errors, expected_errors)
 
+    def test_165_no_raises_unlink(self):
+        extra_params = ["--disable=all", "--enable=no-raise-unlink"]
+        self.assertDictEqual(
+            self.run_pylint([os.path.join(self.root_path_modules, "eleven_module")], extra_params).linter.stats.by_msg,
+            {"no-raise-unlink": 2},
```
luisg123v, line 371: Is this for v11?

Take into account this lint only applies for version 15.0 and above.

luisg123v (changes requested):

luisg123v:
FYI @antonag32, you can close issues from other repositories by using:

`Closes OCA/odoo-pre-commit-hooks#73`

src/pylint_odoo/checkers/function_checker.py
antonag32: Parents are recursively checked to find the enclosing function, for example:

```python
def unlink(self):
    if self.business_logic():
        raise ValueError()  #  offending node, will be reviewed in visit_raise
```

In that case the initial node will be the `raise` but we need to know if it is inside an unlink function so the solution I came up with was to go up the parent tree (effectively up in terms of code lines) until a function definition is found (or the lookup limit is reached).

tests/test_main.py
antonag32: I moved it out to test_module which still has a 10.0 manifest but it seems like that is not very well respected, still may seem less confusing since it has no version in its name.

Also added more code to the test case to ensure this check only runs on 15.0 and above.

antonag32:
Thanks for the review, I made the requested changes, can you check @luisg123v

src/pylint_odoo/checkers/function_checker.py
```diff
@@ -0,0 +1,45 @@
...
+        "no-raise-unlink",
+        "Use @api.ondelete to add any constraints instead",
+    ),
+}
+
+
+class FunctionChecker(OdooBaseChecker):
+    name = "odoolint"
+    msgs = ODOO_MSGS
+    checks_maxmin_odoo_version = {"no-raise-unlink": {"odoo_minversion": "15.0"}}
```
luisg123v, line 19: Could you split this line, please? So it will be easier to add future lints or more conditions.

```suggestion
    checks_maxmin_odoo_version = {
        "no-raise-unlink": {
            "odoo_minversion": "15.0",
        },
    }
```

Besides, it's more readable.

src/pylint_odoo/checkers/function_checker.py
```diff
@@ -0,0 +1,45 @@
...
+            and ("_name" in parent.locals or "_inherit" in parent.locals)
+            and node.name == "unlink"
+        )
+
+    @staticmethod
+    def get_enclosing_function(node, depth=10):
+        parent = getattr(node, "parent", False)
+        for _i in range(depth):
+            if not parent or isinstance(parent, FunctionDef):
+                return parent
```
luisg123v, line 35: I think we could just break and return parent on any case after this loop, IMHO it's clearer that way.

luisg123v (changes requested):

luisg123v:
The following is more of a suggestion, but:

I think closes statements fit better at the bottom, because they works like directives, while the commit description is the actual meesage one expects to read.

antonag32:
I applied the requested changes. Also thought about it and adding an extra file did not seem to make much sense anymore. I added that file thinking the checker would store some state (functions that raise exceptions) and that it was clearer that way. However the checker no longer stores any state and it is actually simple, so I don't think it grants a new file.

Also made some docstring fixes/refactoring around the original code where this new check was added.

luisg123v (approved):
LGTM :+1:

PR #470, opened 2023-09-13
[IMP] new `deprecated-odoo-model-method` check

### LINKED ISSUE #469: Warn/Error on use of fields_view_get in linting Odoo v16.
**Is your feature request related to a problem?**
In v16 `fields_view_get` and associated methods are deprecated. This means that any module that is overriding it is not called. I got caught in a private module migration, but a grep shows that these have been passing through OCA review without question.

**Describe the solution you'd like**
A warning with recommendation to use `get_view`  or `_get_view` as appropriate.

Also note, while the arguments are basically the same, the results of get view are a little different to the old fields_view_get, returning a single arch and no dict of subviews.

antonag32:
This new check looks for model methods that have been deprecated and should not be overriden in model classes. It takes into account odoo versions and comes with sane defaults which can be replaced by a custom list in the configuration file.

Closes #469.

### DIFF
--- a/src/pylint_odoo/checkers/odoo_addons.py
+++ b/src/pylint_odoo/checkers/odoo_addons.py
@@ -107,6 +107,7 @@ from collections import Counter, defaultdict
 import validators
 from astroid import ClassDef, FunctionDef, NodeNG, nodes
 from pylint.checkers import BaseChecker, utils
+from pylint.lint import PyLinter
 
 from .. import misc
 from .odoo_base_checker import OdooBaseChecker
@@ -251,6 +252,11 @@ ODOO_MSGS = {
         "bad-builtin-groupby",
         CHECK_DESCRIPTION,
     ),
+    "W8160": (
+        "%s has been deprecated by Odoo. Please look for alternatives.",
+        "deprecated-odoo-model-method",
+        CHECK_DESCRIPTION,
+    ),
 }
 
 DFTL_MANIFEST_REQUIRED_KEYS = ["license"]
@@ -346,6 +352,8 @@ DFTL_EXTERNAL_REQUEST_TIMEOUT_METHODS = [
     "suds.client.Client",
     "urllib.request.urlopen",
 ]
+DFTL_DEPRECATED_ODOO_MODEL_METHODS = {"16.0": {"fields_view_get"}}
+
 # Regex used from https://github.com/translate/translate/blob/9de0d72437/translate/filters/checks.py#L50-L62  # noqa
 PRINTF_PATTERN = re.compile(
     r"""
@@ -517,6 +525,15 @@ class OdooAddons(OdooBaseChecker, BaseChecker):
                 "help": "List of valid odoo versions separated by a comma.",
             },
         ),
+        (
+            "deprecated-odoo-model-methods",
+            {
+                "type": "string",
+                "metavar": "python_expression",
+                "default": "",
+                "help": "Dictionary consisting of versions (keys) and methods that have been marked as deprecated.",
+            },
+        ),
     )
 
     checks_maxmin_odoo_version = {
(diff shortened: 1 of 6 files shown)


antonag32:
Marked as draft as I still need to write a unit test to verify the options work.

antonag32:
Also need to check the unittests...

src/pylint_odoo/checkers/odoo_addons.py
```diff
@@ -517,6 +525,15 @@ class OdooAddons(OdooBaseChecker, BaseChecker):
                 "help": "List of valid odoo versions separated by a comma.",
             },
         ),
+        (
+            "deprecated-odoo-model-methods",
+            {
+                "type": "csv",
+                "metavar": "<comma separated values>",
+                "default": set(),
```
moylop260, line 533: We generally use the default value in the config parameter

It is impossible since it depends on the odoo version, right?

So, maybe we should use a way to set the dict in the configuration file

Something similar to:
`deprecated_odoo_model_methods={'16.0':['method61','method62'],'17.0': ['method71','method72']}` or something similar

Also, consider that the method is deprecated not only in the current version set but >= current version set

I mean, this 16.0 method will be deprecated for 17.0 too

So, they should be configured to raise the error for all odoo versions >=16.0

I know it is a good practice for configuration files but we can not have new sections using the same class

But I'm opened to use another way

Consider that pre-commit-vauxoo uses only a configuration file for all odoo versions.

antonag32: A dictionary is already being used:

```python
DFTL_DEPRECATED_ODOO_MODEL_METHODS = {"16.0": {"fields_view_get"}}
```

The value for deprecated methods is set on `open` if there was no config provided. I left the options as a CSV because I don't think most people will use it, I see it as a escape hatch, if you are going out of your way to use it, then you need to type out the whole list of methods to consider.

antonag32:
I corrected it so that deprecated methods apply to <= versions. I also tested setting a dictionary on the rc file and it it seems like it is not supported, it is weirdly parsed as a list.

I personally don't see a problem with letting the option be a csv. As far as pre-commit-vauxoo goes, it should work with the defaults right?

src/pylint_odoo/checkers/odoo_addons.py
moylop260: The global variable is a dict

But using the configuration file it is not a dict

I mean, the following line only support a list of methods but they could be not valid for different versions of Odoo

```python
            self._deprecated_odoo_methods = set(self.linter.config.deprecated_odoo_model_methods)
```

For consistency, IMHO we should use dict for configuration file too

Or using set for global variable but not 2 ways to receive the values

If the user needs to define methods deprecated for 16.0 but not for 15.0 it should be able to be configured from configuration file or using the --parameters

antonag32: I think using set is an implementation detail. I use it because it seems to me like the best data structure for the task, a dictionary or list would be inferior for doing `x is in y` type of operations. 

Regarding configuration, I think the following could make us all happy:

```
[ODOOLINT]
deprecated_odoo_model_methods_versions_exclude = list of versions to exclude
```

This configuration lets you cover the following case:

> If the user needs to define methods deprecated for 16.0 but not for 15.0 it should be able to be configured from configuration file or using the --parameters

You could simply do:

`exclude_deprecated_odoo_model_methods_versions = 15.0`

This will exclude version 15.0 methods from the check, but still include methods from all other `valid_odoo_versions`
We could also add one more setting to exclude specific messages regardless of version, and that could be a list as well.

This way we have the most flexibility while keeping the defaults sane and useful for most users.

antonag32:
@moylop260 The config file now accepts a dictionary and in fact for testing purposes I added a .pylintrc with it that should serve as an example for users.

moylop260:
Awesome!

👍

Issue #476, opened 2023-11-14
Pylint-odoo not working from v8.0.6


celm1990:
## Describe the bug
I have a module  with some inconsistencies, but `pre-commit` with `pylint-odoo` v8.0.6(to lastest version) not fails. I try various versions and to `v8.0.5` working fine

## To Reproduce
**Affected versions**:
v8.0.6 +

**Steps to reproduce the behavior:**
1. Create a new module with some inconsistences(I generate repo template from `oca-addons-repo-template` and this use `v8.0.19` for Odoo 17.0 see [here](https://github.com/OCA/oca-addons-repo-template/blob/747209bc359eb6de73e31271a2fac388c9cdc888/src/.pre-commit-config.yaml.jinja#L52)
2. execute `pre-commit run -a` or try `git commit`

**Expected behavior**
`pylint-odoo` must be fails because:
- The manifest version is wrong(16.0 instead of 17.0)
- Text untranslated

**Additional context**
Please see repo where pre-commit not fail(pylint-odoo with lastest version)
https://github.com/celm1990/repo_with_ruff/actions/runs/6841841712/job/18602672369

And this log i change pylint-odoo to v8.0.5 and now working and pre-commit fails
https://github.com/celm1990/repo_with_ruff/actions/runs/6842135353/job/18603271696

I not sure if after [this commit](https://github.com/OCA/pylint-odoo/commit/e0df7c62031d201a1df1a314b65d54b9e67d4a30) is need a extra args on pre-commit to pylint-odoo?

antonag32:
Thanks for reporting. This is happening because `17.0` has not been added yet to pylint-odoo as a valid odoo version. I will create the PR to do so, once added it should work (already tested locally).

celm1990:
Thank you for your reply. I have created a PR at https://github.com/OCA/pylint-odoo/pull/477 This is my first contribution here. Please let me know if anything else is needed.

antonag32:
> Thank you for your reply. I have created a PR at #477 This is my first contribution here. Please let me know if anything else is needed.

I created a PR as well which is basically the same as yours. Only thing missing in yours is running `tox -e update-readme`, to automatically update the readme (that will fix the CI). Try that out and once your MR goes green I can close mine and get your first contribution going.

antonag32:
Also, with this diff you should be able to test the changes before we merge them:

```diff
diff --git a/.pre-commit-config.yaml b/.pre-commit-config.yaml
[...]

PR #477, opened 2023-11-17
[IMP] add support to Odoo v17.0

celm1990:


celm1990:
fixed https://github.com/OCA/pylint-odoo/issues/476

@antonag32 can you review please?

pedrobaeza (approved):
Should we preventively add newer versions?

pedrobaeza:
@sbidoul I think this means a release of the library should be done and we should change the pre-commit config of all of the 17.0 branches.

sbidoul:
@pedrobaeza I don't understand. This is PR just changes the default value for the list of allowed versions, and on 17.0 branches we have the correct value: https://github.com/OCA/mis-builder/blob/cf03aa9a706b9cb678ab4fd57f901852e4df007e/.pylintrc#L13

pedrobaeza:
Please check the origin issue. It seems the checks are not being done correctly on current pylint-odoo version for 17.0.

sbidoul:
Yeah I know, but I doubt this PR would fix that.

pedrobaeza:
Well, according one Vauxoo's guy in #478, it seems this is the way to go.

celm1990:
> Yeah I know, but I doubt this PR would fix that.

On my repo for example, I changed `pre-commit` configuration  and now this working

image

antonag32:
> @antonag32
> 
> Could you check why this PR is fixing the error even if the pylint.cfg has the value explicitly assigned, please?
> 
> https://github.com/celm1990/repo_with_ruff/blob/c46854ef3685bbd413277f22e6fb0dc72382484a/.pylintrc-mandatory#L12
> 
> Even if we should support the new version by default, the explicitly value assigned should work too as @sbidoul said

It is this piece of code:
https://github.com/OCA/pylint-odoo/blob/9d85855b850d23d9004ac27b2e128b678888a19d/src/pylint_odoo/checkers/odoo_base_checker.py#L35-L40

The condition `<= misc.version_parse(odoo_maxversion)` is not met since odoo_maxversion is `16.0` as of now, and odoo_version is set as `17.0` on the config file.

Because `is_odoo_message_enabled` gets called when adding new messages:
https://github.com/OCA/pylint-odoo/blob/9d85855b850d23d9004ac27b2e128b678888a19d/src/pylint_odoo/checkers/odoo_base_checker.py#L42-L46

No messages will ever get added.

antonag32:
I wen't with adding `17.0` because it was a simple fix which fits with the current program logic. We could look into making pylint-odoo future proof and not require this sort of changes (adding new odoo versions), but I don't know if the project is interested in that (cost-benefit). 

I looked at `is_odoo_message_enabled` and could not think of a fix to it that was as simple as the one in this PR. Maybe someone else can help.

moylop260:
@antonag32 

Ok, thank you

Let's to do the quick fix then

moylop260:
FYI it will be released to https://pypi.org/project/pylint-odoo/9.0.2 after the following build finishes https://github.com/OCA/pylint-odoo/actions/runs/6906205867/job/18790705636

pedrobaeza:
So we need to update all 17.0 branches pylint-odoo version, isn't it?

moylop260:
> So we need to update all 17.0 branches pylint-odoo version, isn't it?

Hi @pedrobaeza 

Yes, please

pedrobaeza:
@sbidoul you have the tools for such massive upgrade... I can do the PR to oca-addons-template though.

sbidoul:
> @sbidoul you have the tools for such massive upgrade... I can do the PR to oca-addons-template though.

Will do. No urgency though, right?

pedrobaeza:
No critical checks indeed.

celm1990:
for this PR https://github.com/Tecnativa/doodba-copier-template/pull/425 is need upgrade https://github.com/OCA/oca-addons-repo-template/pull/235 to pass test

pedrobaeza:
@celm1990 I don't see why we need the OCA addons template for Doodba. Just put the new pylint-odoo version in Doodba copier template.

celm1990:
> @celm1990 I don't see why we need the OCA addons template for Doodba. Just put the new pylint-odoo version in Doodba copier template.

@pedrobaeza because [this test](https://github.com/Tecnativa/doodba-copier-template/blob/b871898d2fcb9052b7961cec89a63aa968c130cd/tests/test_nitpicking.py#L492-L504) change version module and expected raise error, but `pylint-odoo` not raise error. Or did I misunderstand?

pedrobaeza:
But as said, just bump the pylint-odoo version to 9.0.2 in Doodba.

PR #481, opened 2023-12-15
[FIX] odoo-addons-relative-import: exclude migrations

### LINKED ISSUE #480: odoo-addons-relative-import: Ignore files under the migrations directory
Since migration scripts are not actually Python packages (because there are no `__init__.py` files at the same level as them), the lint `odoo-addons-relative-import` should not be triggered, because using relative imports from migration scripts is not possible.

CC @antonag32 @moylop260  @xmglord

antonag32:
Since migration files don't have an __init__,py, they can't perform relative imports, leaving absolute imports as the only option. Therefore this message does not apply to them.

Fixes #480.


antonag32:
Tested with kukyflor by changing `.pre-commit-config-optional.yaml` to use a local pylint_odoo installation (local system hook)

PR #485, opened 2024-02-12
[ADD] prohibited-method-override: Add a new check to identify methods marked as prohibited to override

Pexers:
# Pull Request Template

## Description
- Check for prohibited methods being overridden.
- Enables the user to specify, for example, core Odoo methods that should remain unaltered and cannot be overridden under any circumstances.

## Type of change

Please delete options that are not relevant.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## How Has This Been Tested?

1. There's a new test named `test_175_prohibited_method_override` within _test_main.py_.
2. Run the command below to start the test (using Python v3.9).
```sh
$ tox -e py39  -- -k test_175_prohibited_method_override
```
3. Verify that the test passes with two methods prohibited from being overridden. ✅ 
```sh
tests/test_main.py::MainTest::test_175_prohibited_method_override ************* Module broken_module.tests.test_model
testing/resources/test_repo/broken_module/tests/test_model.py:15 Prohibited override of "some_base_method_1" method. - [prohibited-method-override]
testing/resources/test_repo/broken_module/tests/test_model.py:15 Prohibited override of "some_base_method_2" method. - [prohibited-method-override]

PASSED
```

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] Any dependent changes have been merged and published in downstream modules
- [x] I have checked my code and corrected any misspellings

### DIFF
--- a/src/pylint_odoo/checkers/odoo_addons.py
+++ b/src/pylint_odoo/checkers/odoo_addons.py
@@ -204,6 +204,7 @@ ODOO_MSGS = {
     "W8103": ('Translation method _("string") in fields is not necessary.', "translation-field", CHECK_DESCRIPTION),
     "W8105": ('attribute "%s" deprecated', "attribute-deprecated", CHECK_DESCRIPTION),
     "W8106": ('Missing `super` call in "%s" method.', "method-required-super", CHECK_DESCRIPTION),
+    "W8107": ('Prohibited override of "%s" method.', "prohibited-method-override", CHECK_DESCRIPTION),
     "W8110": ("Missing `return` (`super` is used) in method %s.", "missing-return", CHECK_DESCRIPTION),
     "W8111": (
         'Field parameter "%s" is no longer supported. Use "%s" instead.',
@@ -303,6 +304,7 @@ DFTL_METHOD_REQUIRED_SUPER = [
     "unlink",
     "write",
 ]
+DFTL_PROHIBITED_OVERRIDE_METHODS = []
 DFTL_CURSOR_EXPR = [
     "cr",  # old api
     "self._cr",  # new api
@@ -493,6 +495,15 @@ class OdooAddons(OdooBaseChecker, BaseChecker):
                 "metavar": "<comma separated values>",
                 "default": DFTL_METHOD_REQUIRED_SUPER,
                 "help": "List of methods where call to `super` is required.separated by a comma.",
+            },
+        ),
+        (
+            "prohibited-method-override",
+            {
+                "type": "csv",
+                "metavar": "<comma separated values>",
+                "default": DFTL_PROHIBITED_OVERRIDE_METHODS,
+                "help": "List of methods that have been marked as prohibited to override.",
             },
         ),
         (
@@ -1188,7 +1199,9 @@ class OdooAddons(OdooBaseChecker, BaseChecker):
 
         return node.name in self._deprecated_odoo_methods
 
-    @utils.only_required_for_messages("method-required-super", "missing-return", "deprecated-odoo-model-method")
+    @utils.only_required_for_messages(
+        "method-required-super", "prohibited-method-override", "missing-return", "deprecated-odoo-model-method"
+    )
     def visit_functiondef(self, node):
         """Check that `api.one` and `api.multi` decorators not exists together
         Check that method `copy` exists `api.one` decorator
@@ -1215,6 +1228,22 @@ class OdooAddons(OdooBaseChecker, BaseChecker):
             if isinstance(func, nodes.Name) and func.name == "super":
                 there_is_super = True
                 break
+
+        # Verify if super attributes are prohibited methods to override
+        if there_is_super and self.linter.config.prohibited_method_override or DFTL_PROHIBITED_OVERRIDE_METHODS:
+            for attr in node.nodes_of_class(nodes.Attribute):
+                if attr.attrname != node.name:
+                    continue
+                func = attr.expr.func
+                if (
+                    isinstance(func, nodes.Name)
+                    and func.name == "super"
+                    and (
+                        attr.attrname in self.linter.config.prohibited_method_override
+                        or attr.attrname in DFTL_PROHIBITED_OVERRIDE_METHODS
+                    )
+                ):
+                    self.add_message("prohibited-method-override", node=node, args=(attr.attrname,))
 
         there_is_return = any(node.nodes_of_class(nodes.Return, skip_klass=(nodes.FunctionDef, nodes.ClassDef)))
         if (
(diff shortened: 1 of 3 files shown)


antonag32:
Thanks for the PR. I briefly went over your implementation and I have some questions.

It seems like your current code looks for super() calls to attributes in the prohibited override methods variable. If that is the case, 
it does not take into account the method where it is being called. For example, is this overwriting anything?

```python3
def my_method(self):
    super().some_other_method()
```

It seems to me that an overwrite would look more like this:
```python3
def my_method(self):
    super().my_method()
```

I also don't have merging rights but I think this feature is too user specific, so maybe it makes sense to develop it as a separate plugin.

Pexers:
Hey @antonag32, first and foremost, thanks for the quick reply! 😄 

I've made a small change to fix the raised issue concerning attribute & parent method names not being equal.

In regards to this check being too user specific, I understand that overwriting Odoo core methods is part of development workflow. However, I would like to raise awareness among developers within my organization that there are some base methods considered to be sensitive, and extra care should be taken when modifying them.

Nevertheless, this check won't raise any errors, only warnings, and it will remain optional. 🙂

sconetto:
hello @antonag32 @Pexers,

I believe this is a very valid addition, we had some quality issues of engineers overriding base Odoo methods without the proper care of managing super calls. I feel this would be an awesome thing to had in the pylint-odoo.

👍🏻 +1

yajo (approved):
LGTM, although I can't imagine a use case for this 🤷🏼‍♂️

testing/resources/test_repo/broken_module/tests/test_model.py
```diff
@@ -11,3 +11,15 @@ def method1(self, example_var):
...
+        # Should fail
+        super().test_base_method_1()
+        # Should not fail
+        return super().test_base_method_2()
+
+    def test_base_method_2(self):
+        # Should not fail
+        super().test_base_method_1()
+        # Should fail
+        return super().test_base_method_2()
```
luisg123v, line 25: There are two cases that should fail and two that should not. For safety reasons, I think they should be different, to ensure we're catching the expected ones.

testing/resources/test_repo/broken_module/tests/test_model.py
```diff
@@ -11,3 +11,15 @@ def method1(self, example_var):
...
+    
+    def test_base_method_1(self):
+        # Should fail
+        super().test_base_method_1()
+        # Should not fail
+        return super().test_base_method_2()
+
+    def test_base_method_2(self):
+        # Should not fail
+        super().test_base_method_1()
```
luisg123v, line 23: Could you add cases using the class name, please? To check if those also work.

I mean:

`super(TestModel, self).test_base_method_1()`

luisg123v (changes requested):
Could you squash commits and provide a meaningful commit description, please?

testing/resources/test_repo/broken_module/tests/test_model.py
Pexers: Added ✅ They do work, thank you!

testing/resources/test_repo/broken_module/tests/test_model.py
Pexers: Let me know if the latest changes meet with your suggestion. 😄 Thank you.

src/pylint_odoo/checkers/odoo_addons.py
```diff
@@ -1216,6 +1229,24 @@ def visit_functiondef(self, node):
                 there_is_super = True
                 break
 
+        # Verify if super attributes are prohibited methods to override
+        if there_is_super and (
+            self.linter.config.prohibited_method_override or len(DFTL_PROHIBITED_OVERRIDE_METHODS) != 0
```
luisg123v, line 1234: This is a lint in more recent versions, because:

`if x != 0:`' 

May be simplified as:

`if x:`

```suggestion
            self.linter.config.prohibited_method_override or DFTL_PROHIBITED_OVERRIDE_METHODS
```

luisg123v (changes requested):

luisg123v:
@Pexers the following was not addressed:

> Could you squash commits and provide a meaningful commit description, please?

I think there should be only one commit.

testing/resources/test_repo/broken_module/tests/test_model.py
luisg123v: Are you sure this is addressed?

I mean, there are four tested cases, half of them should faild and the other half should not. If lint logic is inverted and the ones that shouldn't fail are actually failing and vice versa, we wouldn't notice it.

Maybe we could add more cases, e.g. the ones you added with the old super syntax but passing contexts or similar.

We usually try to address as many combinations as possible, to ensure there aren't false negatives or false positives.

Another improvement could be to call three different methods and prohibit two of them.

Pexers: I understand your point now, thanks for the clarification. 👍

I've decided to follow your suggestion and prohibit the override of two out of three different methods to prevent the described problem from happening.

src/pylint_odoo/checkers/odoo_addons.py
```diff
@@ -1216,6 +1229,24 @@ def visit_functiondef(self, node):
                 there_is_super = True
                 break
 
+        # Verify if super attributes are prohibited methods to override
+        if there_is_super and (
+            self.linter.config.prohibited_method_override or DFTL_PROHIBITED_OVERRIDE_METHODS
```
luisg123v, line 1234: This line is causing one of the linters to fail (black). Could you fix it, please?

You just need to replace it by:

`if there_is_super and self.linter.config.prohibited_method_override or DFTL_PROHIBITED_OVERRIDE_METHODS:`

luisg123v (changes requested):
We're almost done!

Just a couple of suggestions regarding commit message:
- It includes PR number, but it shouldn't
- Title is too long (103 characters). Commit titles are usually stripped to 72 characters. I'd recommend summarizing, e.g.:
  
`[ADD] prohibited-method-override: new check to avoid overriding methods`

Pexers:
All done ✅ Thank you for your time guys! 😄

luisg123v (approved):
@Pexers thanks for contributing!

LGTM :+1: 

@moylop260 Could you merge, please?

PR #502, opened 2024-12-18
[IMP] manifest-behind-migrations: Improve detection of the check only using only odoo manifest file parsing

moylop260:


src/pylint_odoo/checkers/odoo_addons.py
```diff
@@ -106,6 +106,7 @@
 
 import validators
 from astroid import ClassDef, FunctionDef, NodeNG, nodes
+from packaging.version import Version
```
antonag32, line 109: I don't think this is even required? Apparently `max` works just fine (already being used for comparing migration versions) as it returns the lexicographically largest string, so it would be neat if we can just use inbuilt max.

moylop260: If you create a folder `migrations/no_version_format`  the max result will be weird

`>> max("0.1", "no_version_format")`
`Result: no_version_format`

The `packaging.version.Version` raises:
 - `packaging.version.InvalidVersion: Invalid version: 'no_version_format'`

PR #512, opened 2024-12-20
DRAFT: [FIX] consider-merging-classes-inherited: Fix corner case using jobs

moylop260:
It is even too complex than expected since that there are new behaviours using `jobs`

See the following slow case ~10m:
 - https://gist.github.com/moylop260/1627cc4585ff790edebcb93bd68bdc08


moylop260:
I will re-open this MR when I realized what is the issue related with jobs duplicating the results

It could be related to current pylint version but not sure

PR #525, opened 2025-03-04
[IMP] add manifest-external-assets check

antonag32:
This new check verifies that no external sources are referenced on modules' assets.

Related to: https://github.com/OCA/odoo-pre-commit-hooks/issues/124.

### DIFF
--- a/src/pylint_odoo/checkers/odoo_addons.py
+++ b/src/pylint_odoo/checkers/odoo_addons.py
@@ -104,6 +104,7 @@ import re
 import string
 import warnings
 from collections import Counter, defaultdict
+from urllib.parse import urlparse
 
 from astroid import ClassDef, FunctionDef, NodeNG, nodes
 from pylint.checkers import BaseChecker, utils
@@ -267,6 +268,11 @@ ODOO_MSGS = {
     "W8161": (
         "Better using self.env._ More info at https://github.com/odoo/odoo/pull/174844",
         "prefer-env-translation",
+        CHECK_DESCRIPTION,
+    ),
+    "W8162": (
+        "Asset %s should be distributed with module's source code. More info at https://httptoolkit.com/blog/public-cdn-risks/",
+        "manifest-external-assets",
         CHECK_DESCRIPTION,
     ),
 }
@@ -1042,6 +1048,7 @@ class OdooAddons(OdooBaseChecker, BaseChecker):
         "resource-not-exist",
         "website-manifest-key-not-valid-uri",
         "manifest-behind-migrations",
+        "manifest-external-assets",
     )
     def visit_dict(self, node):
         if not os.path.basename(self.linter.current_file) in misc.MANIFEST_FILES or not isinstance(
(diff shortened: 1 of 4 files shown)


antonag32:
Can you review @moylop260 @luisg123v?

src/pylint_odoo/checkers/odoo_addons.py
```diff
@@ -1197,6 +1204,32 @@ def visit_dict(self, node):
...
+    def _check_manifest_external_assets(self, node):
+        def is_external_url(url):
+            return urlparse(url).scheme
+
+        for _, item in node.items:
+            for element in item.elts:
+                if isinstance(element, nodes.Const):
+                    if is_external_url(element.value):
+                        self.add_message("manifest-external-assets", node=element)
+                elif isinstance(element, nodes.Tuple):
```
moylop260, line 1228: Notice it could be a Tuple or a List

See the following example:
 - https://github.com/odoo/odoo/blob/106dc30bebf49335eb7d2450a55d14cd46469e17/addons/barcodes/__manifest__.py#L18

```suggestion
                elif isinstance(element, (nodes.Tuple, nodes.List)):
```

src/pylint_odoo/checkers/odoo_addons.py
```diff
@@ -1197,6 +1204,32 @@ def visit_dict(self, node):
...
             self.add_message("manifest-maintainers-list", node=manifest_keys_nodes.get("maintainers") or node)
 
+        # Check there are no external assets
+        if self.linter.is_message_enabled("manifest-external-assets"):
+            assets_node = None
+            for item in node.items:
+                if item[0].value == "assets":
+                    assets_node = item[1]
+
+            # it is important to use the actual astroid.node instead of manifest_dict, otherwise the
```
moylop260, line 1214: I would prefer to use `manifest_dict` + `args` in order to use the string value with the problem

I mean,

Currenly it is good showing the number of line:

```
twelve_module/__manifest__.py:15:12: {msg}
twelve_module/__manifest__.py:19:12: {msg}
twelve_module/__manifest__.py:20:12: {msg}
```


But it could be good too using 
```
twelve_module/__manifest__.py:12:14: {msg} {wrong url}
twelve_module/__manifest__.py:12:14: {msg} {wrong url}
twelve_module/__manifest__.py:12:14: {msg} {wrong url}
```

So, the result final could be:

```txt
twelve_module/__manifest__.py:12:14: W8162: Assets https://shady.cdn.com/somefile.js should be distributed with module's source code. More info at https://httptoolkit.com/blog/public-cdn-risks/ (manifest-external-assets)
twelve_module/__manifest__.py:12:14: W8162: Assets https://bad.idea.com/cool.css should be distributed with module's source code. More info at https://httptoolkit.com/blog/public-cdn-risks/ (manifest-external-assets)
```

It should be good since that manifest_dict is more pythonistic than using astroid nodes

So, it is easy to maintain

Only consider a recent fix that I have added for values wrong formed

src/pylint_odoo/checkers/odoo_addons.py
moylop260: In fact, Using manifest_get you could get all the isinstance(..., iterable) object if it is not a string and call it recursively

src/pylint_odoo/checkers/odoo_addons.py
antonag32: I originally tried using manifest_dict but had issues with test_180 which runs parallel jobs. It seems like message were being mangled up because all errors pointed to the same root node (the assets node), also the linenumber was wrong since we were pointing to assets node, instead of the actual offending URL.

src/pylint_odoo/checkers/odoo_addons.py
antonag32: I was checking if it was a tuple because of this use case:

https://github.com/odoo/odoo/blob/032ee36a9556688e2f022ddfb2081800b80efef1/addons/website/__manifest__.py#L241

Odoo docs always show this usage with tuples but I guess in theory you could also use a List although I have not seen it before, I will add it so we cover all possibilities.

src/pylint_odoo/checkers/odoo_addons.py
moylop260: Good point for parallels!!!

I didn't know it

I have bypassed a few checks for parallels

Could you check if you can fix them (another PR)

moylop260: Forget it

I think it is a different matter

PR #529, opened 2025-05-13
[ADD] inheritable-method-lambda: Add check to be inheritable the domain and default using lambda

moylop260:
[REF] string-method-name: Rename check to inheritable-method-string

Fix https://github.com/OCA/odoo-pre-commit-hooks/issues/126


luisg123v (approved):
LGTM :+1:

PR #546, opened 2025-09-26
[ADD] super-method-mismatch: Check super() calls match current method

moylop260:
Detects mismatches between the current method
and the method called via super()

e.g.

    def name_get(self):
        return super().name_get2()

This rule ensures that `super()` calls always reference the same
method being defined, preventing mismatches


luistorresm (approved):
LGTM @moylop260

PR #459, opened 2023-05-24
[ADD] checkers: category-allowed

ThomasBinsfeld:

PR #547, opened 2025-09-26
[REF] category-allowed: Restrict categories to match Odoo Apps filters (only for apps)

moylop260:
Odoo Apps (https://apps.odoo.com/apps) only supports a limited set of categories,
matched exactly using the `=` operator.
    - apps
    - Categories not listed there will not be shown.
    - Nested categories (e.g. category/subcategory) are also ignored.

To ensure apps are correctly filtered and displayed under the intended category,

It introduces a new section that restricts the allowed categories only for apps
(modules with `price` key in the manifest)

Continuation of https://github.com/OCA/pylint-odoo/pull/459


pedrobaeza:
You may want to create new categories because your modules expand these possible applications, so I don't think this is a good idea.

luistorresm (approved):
LGTM @moylop260 

Thanks 👍

PR #575, opened 2026-07-03
Revert "[ADD] translation-injection: Detect str.format() inside Odoo translation methods

moylop260:
This reverts commit https://github.com/OCA/pylint-odoo/commit/e6ca8f245659782a7c6a88affd3acd6d40d0ced8.

Odoo has the following moneky patch to avoid calling unsafe format sentences:
 - https://github.com/odoo/odoo/blob/8ee78c39f48b2f501a45cdfffca76b23d8656a51/odoo/_monkeypatches/_cpython.py#L71

So, this check doesn't have sense for Odoo

It was a reply from security email from Odoo team

I double confirmed using `odoo-bin shell` vs python3
 - Screenshot 2026-07-03 at 4 42 06 p m

So, the security issue is not reproduced from Odoo


SirPyTech (changes requested):
Thanks for looking more deeply into this!
I have checked too and can confirm that the monkeypatch from Odoo prevents from accessing some _unsafe_ attributes like `__class__`.

I have done some more tests in the `odoo shell`:
```py
env: <odoo.api.Environment object at 0x7fca0345a050>
odoo: <module 'odoo' from '/opt/odoo/custom/src/odoo/odoo/__init__.py'>
openerp: <module 'odoo' from '/opt/odoo/custom/src/odoo/odoo/__init__.py'>
self: res.users(1,)
Python 3.10.20 (main, Jun 11 2026, 01:13:00) [GCC 12.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
(Console)
>>>
```

and even the patched `format` allows accessing attributes, so something like:
```py
>>> "My user is {user.login}".format(user=self)
'My user is __system__'
```
would allow a translator to change the string to access other fields of the record, like:
```py
>>> "My user is {user.id}".format(user=self)
'My user is 1'
```
or the `Environment`:
```py
>>> "My user is {user.env}".format(user=self)
'My user is <odoo.api.Environment object at 0x7fca0345a050>'
```
even if this environment is somehow safer because it does not allow `browse` or `search`.

I honestly can't evaluate the level of security risk; I know this is much less severe than accessing `__globals__`, but this should probably be not allowed so I'd keep the `translation-injection` warning.

Just for comparison, using the good old `%` interpolation is safe:
```py
>>> "My user is %(user.login)s}" % dict(user=self)
  File "<console>", line 1, in <module>
KeyError: 'user.login'
```

moylop260:
I think You have a better point here

Thank you!

PR #576, opened 2026-07-13
[FIX] invalid-name: false positive for executable with .py extension

luisg123v:
Currently a file is exempted from `invalid-name` only when all of these conditions are met:
- its first line is a shebang
- it has the executable bit set
- its extension is not `.py`

However, a file with a shebang and executable permissions is a script, regardless of its extension. The `.py` extension may still be kept to make it clear that the script is written in Python or to ensure that Python linters are applied to it.

To fix this, remove the `.py` condition so any executable module with a shebang is exempt.


luisg123v:
Hi @moylop260,

Could you review, please?

CC @desdelinux

moylop260:
@luisg123v 

IMHO, if you are using a `shebang` comment and executable mode `chmod +x`
so the extension is noise since you should be able to run the file without knowing (or caring) what language it's written in

In fact, using an executable file without an extension will give you the freedom to change the language without changing the extension

We had this case a long time ago from the MQT project:
 - Notice the project was initialized using bash scripts:
   - [test_pylint (bash)](https://github.com/OCA/maintainer-quality-tools/blob/267214cbdc3d390d045c6f8a040f98b0b7b021d5/travis/test_pylint)
 - After the scripts were changed to Python:
   - [test_pylint (py)](https://github.com/OCA/maintainer-quality-tools/blob/9f1fe61c562fc572937df4a03151b8a442d66d5e/travis/test_pylint)

Notice that the name of the file was not changed
Also, the extension was not used

Otherwise, we should change all the projects using the old extension `test_pylint.bash` to be changed to `test_pylint.py`

But this change was transparent for all the projects since that it was a change in the content of the file and shebang comment without extension noise

luisg123v:
@moylop260 

> IMHO, if you are using a `shebang` comment and executable mode `chmod +x` so the extension is noise since you should be able to run the file without knowing (or caring) what language it's written in

It's not noise, some tools don't work properly if the extension is not explicit. For instance:
- black
- prettier
- pylint

These work under `pre-commit` because the files are classified before being passed to the tools. But, if you invoke any of the above tools recursively, all will skip files without extension. Only `isort` will catch them.

Other legitimate reasons to prefer the extension:
- Globs *.py
- Readability inside one directory with several scripts in different languages.

> In fact, using an executable file without an extension will give you the freedom to change the language without changing the extension

I'd say this is a preference, not a rule. There are valid reasons for preferring one or the other. The problem with the current approach is, it forces to go one and discard the other.

Regards,

moylop260 (approved):
👍 

I think you have better arguments

PR #578, opened 2026-07-18
[ADD] translation-markup-unsafe-xss: Detect unsafe Markup translation

moylop260:
Add translation-markup-unsafe-xss to flag
`Markup(self.env._(..., kwargs))` patterns,
where translation happens in Markup
and it marks the string as safe.

This prevents missing XSS-prone constructions while keeping
the safe Markup(...) % values pattern accepted.

Even if it is inspired by Bandit [B704 check](https://bandit.readthedocs.io/en/latest/plugins/b704_markupsafe_markup_xss.html)
    - Potential XSS with ``markupsafe.Markup`` detected.
    - Do not use ``Markup`` on untrusted data. - [B704] - (MEDIUM)

It is better to have a custom message for translation case
since that we have a lot of lint related to translations
and it could be confuse for developers trying to fix a translation

Kudos @randall-vx


moylop260:
Better using https://bandit.readthedocs.io/en/latest/plugins/b704_markupsafe_markup_xss.html 
or better S704 unsafe-markup-use from ruff faster

PR #587, opened 2026-08-19
[ADD] release: Cut a release from a button without a laptop

moylop260:
Releasing needs `bump2version` run from a checkout, so it cannot be done without a particular laptop. This runs the same commands from the GitHub UI instead — one button. **No new secrets, no personal access token.**

## What it does

`workflow_dispatch` only, with a `part` input (`patch` / `minor` / `major`, default `patch`), guarded so it only ever acts on `OCA/pylint-odoo` and only from `main`, and serialized with `concurrency: release`.

1. Checkout, python, dependencies.
2. **Refuse to release twice** — aborts when the last commit of `main` is already a `Bump version:` commit, which is what both a second run of this workflow and a release already cut from a checkout leave behind.
3. **Validate** with `tox -e build`, the same checks the tag will run, so a broken build never becomes a published version. It also dry-runs bump2version, confirming the search patterns still match the files to patch.
4. **Release**: `bump2version <part> --no-sign-tags` and `git push origin main --tags`, the same commands used from a checkout.
5. **Ask `build` to publish** with `gh workflow run test.yml --ref v<version>`.

The upload itself is not done here: `build` publishes to PyPI with the existing `PYPI_API_TOKEN`, exactly as it does today.

## Why step 5 exists

A tag pushed with the `GITHUB_TOKEN` does not emit the push event that starts `build`, so it would sit there unpublished. That is GitHub's anti-recursion rule, and no amount of waiting helps — the run is never created:

> "events triggered by the `GITHUB_TOKEN`, with the exception of `workflow_dispatch` and `repository_dispatch`, will not create a new workflow run" — [docs](https://docs.github.com/en/actions/concepts/security/github_token)

The usual workaround is a personal access token, which for a community organization means tying releases to one person's credentials and their expiry. This uses the documented exception instead: `workflow_dispatch` always creates a run, even with the `GITHUB_TOKEN`, so `build` is requested explicitly.

That needs two small changes in `build`:

- accept `workflow_dispatch`
- the publish step checks only `startsWith(github.ref, 'refs/tags')` instead of also requiring `github.event_name == 'push'`, so a run dispatched on a tag publishes too

A tag pushed from a checkout keeps starting `build` by itself, unchanged.

## Trade-off: the tags are not signed

`sign_tags = True` needs a GPG key that only exists in the maintainer's laptop, so the workflow passes `--no-sign-tags`. Releases cut from the button produce an annotated but unsigned tag and an unsigned commit; `v10.0.11` and older stay signed. The pending `# TODO: Publish package only for signed tags` is therefore unachievable for this path and is replaced by a note describing both publication paths.

## Verification

- `actionlint` (with shellcheck) clean on both workflows.
- The release step was simulated in a throwaway clone configured like a runner (no GPG key, Actions-style identity): it produces the commit `Bump version: 10.0.10 → 10.0.11`, patches the same 4 files as a local release, and creates `v10.0.11` as an annotated unsigned tag.
- The tag name handed to `gh workflow run` is read back from `.bumpversion.cfg` after the bump and verified to match the tag bump2version actually created.
- The double-bump guard was checked against both a real bump subject (aborts) and a normal one (proceeds).
- Fail-safe confirmed by accident during the simulation: when `git tag` fails because the tag already exists, `set -euo pipefail` aborts the step before `git push`, so a half-finished release never reaches the remote.

## Notes

- The bump commit pushed to `main` does not start a CI run either, for the same anti-recursion reason. It only changes version strings and `tox -e build` already ran on that tree, but it is a difference worth knowing.
- `workflow_dispatch` only shows the button once the file is on the default branch, so the first real run is also the first end-to-end test. A `dry_run` input that stops right before `bump2version` would make that first run safe — say the word and I'll add it.


sbidoul:
@moylop260 please try to not rely on a GitHub token obtained from a secret. This should be doable using job tokens with the correct `permissions:`. See for instance https://github.com/pypa/build/blob/main/.github/workflows/cd.yml

PR #589, opened 2026-09-23
[FIX] odoo_addons: Do not ask LazyTranslate to use self.env._

federicobond:
`prefer-env-translation` fired on both names in `TRANSLATION_METHODS`, so `_lt` was reported alongside `_`. 

LazyTranslate defers the lookup to `__str__` instead of returning a string, so `self.env._` is not a replacement for it, and its documented use is the global that has no env to reach.


tests/test_main.py
```diff
@@ -527,6 +527,39 @@ def fstring_no_sqli(self):
...
+class Model(models.Model):
+    def method(self):
+        lazy = _lt("Deferred")
+        good = self.env._("Reported by nothing")
+        bad = _("Reported")
+        return lazy, good, bad
+"""
+        with TemporaryDirectory() as tmp_dir:
+            tmp_path = os.path.join(tmp_dir, "lazy_translate.py")
+            with open(tmp_path, "w", encoding="utf-8") as tmp_f:
```
federicobond: I used this pattern so the test case can run on Windows. Happy to adapt the other tests that are currently skipped on Windows if you want.

